PR #22, opened 2025-12-21
Refactor agent breaking out code review comment skills

theMickster:
## 📔 Objective
Refactored code review agent to use specialized skills, reducing duplication and making the system more maintainable. 
- Created classifying-review-findings skill - Defines severity categories and validation criteria
- Created posting-bitwarden-review-comments skill - Handles comment formatting and GitHub posting protocols
- Registered skills in plugin.json
- Agent now invokes skills for reference material vs. embedding everything inline
- Cleaned up the few places where we used 💭 and instead always use❓
- Took a lot of duplication out of the `README.md` and, instead, pointed uses to where that information lives
- Fixed a number of problems with the DENY list formatting; Claude Code threw errors loading the `settings.json` when debugging locally.
- Started down the path of creating some steps to test our plugin using a `TESTING.md`; not perfect but feels like a good start.

I have found more consistent success with Claude Code CLI finding and using the new skills when I added the following sentences to the `/code-review-local ` slash command prompt. Therefore, I think we need to add these sentences to our `gh-action` prompt to firmly nudge Claude Code to discover our new skills by name.

```
Use the classifying-review-findings skill to classify findings.
Use the posting-bitwarden-review-comments skill to format inline and summary comments.
```

### DIFF
--- a/plugins/bitwarden-code-review/agents/bitwarden-code-reviewer/AGENT.md
+++ b/plugins/bitwarden-code-review/agents/bitwarden-code-reviewer/AGENT.md
@@ -3,7 +3,7 @@ name: bitwarden-code-reviewer
 version: 1.4.0
 description: Specialized agent for conducting thorough, professional code reviews following Bitwarden engineering standards. Focuses on security, correctness, and high-value feedback with minimal noise. Use when reviewing pull requests, analyzing code changes, or when user requests code review feedback. PROACTIVELY invoke when user mentions "review", "PR", or "pull request".
 model: sonnet
-tools: Read, Bash(gh pr view:*), Bash(gh pr diff:*), Bash(gh pr checks:*), Bash(./scripts/get-review-threads.sh:*), Grep, Glob, Skill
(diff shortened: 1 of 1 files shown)


plugins/bitwarden-code-review/agents/bitwarden-code-reviewer/AGENT.md
```diff
@@ -157,7 +157,7 @@ This prevents duplicate comments and maintains conversation continuity.
 - **Jira Reference**: Expected in the `## 🎟️ Tracking` section
 - **Test Plan**: Expected to describe how changes were verified, or reference test plan in linked Jira task
 
-If deficient, create a finding (💭) with rewrite suggestions in a collapsible `<details>` section.
```
withinfocus, line 160: I kinda thought this would be the bubble and not a question.

theMickster: Claude pointed out that it wasn't aligned with the emoji's we had in other places. I chose the question to start with because I feel like those two things lacking can be a place for Claude to ask a question as to whether they are needed or not. We can change it to a 🎨 and have it make a suggestion

PR #43, opened 2026-01-16
Enhance our pr comment thread retrieval process

theMickster:
## 📔 Objective

Updates the bitwarden-code-review plugin to consume pre-fetched PR thread data from /tmp/pr-threads.json when available. This enables reliable context delivery by injecting thread data directly into the agent's prompt rather than relying on the agent to invoke a skill. The change supports both GitHub Actions workflows (pre-fetched threads) and local review (API fallback).

  ---
This pairs with the get-pull-request-threads action added to the gh-actions repo which creates the JSON file before Claude Code runs. [Create composite action to retrieve pull request threads](https://github.com/bitwarden/gh-actions/pull/574)


plugins/bitwarden-code-review/skills/detecting-existing-threads/SKILL.md
```diff
@@ -11,10 +11,27 @@ Prevent duplicate comments by detecting existing review threads before posting n
...
 ## Required Tools
 
-- `Bash(gh pr view:*)` - Get general PR comments
-- `Bash(gh api graphql -f query=:*)` - Get resolved and open review threads (direct GraphQL)
+- `Read` - Read pre-fetched thread data from `/tmp/pr-threads.json`
+- `Bash(gh pr view:*)` - Get general PR comments (fallback)
+- `Bash(gh api graphql -f query=:*)` - Get resolved and open review threads (fallback)
 
-## Step 1: Determine PR Number
+## Step 0: Check for Pre-fetched Thread Data
```
withinfocus: ❓ Will Claude like a "step 0"? Might be better to just renumber.

theMickster:
> Is the script that we had trouble getting Claude to remove also gonna be deleted? I thought we felt it better at this point to simplify both the permissions exclusions and perhaps even removing the thread skill.

The script has been deleted, but I kept the instructions and the Skill to support thread detection for local reviews. 
Would you like to see a more evasive refactor with the skill completely removed and the instructions for retrieving threads for a local review moved into the `code-review-local.md` command?

withinfocus:
> > Is the script that we had trouble getting Claude to remove also gonna be deleted? I thought we felt it better at this point to simplify both the permissions exclusions and perhaps even removing the thread skill.
> 
> The script has been deleted, but I kept the instructions and the Skill to support thread detection for local reviews. Would you like to see a more evasive refactor with the skill completely removed and the instructions for retrieving threads for a local review moved into the `code-review-local.md` command?

I am just skittish about this working or not. Fine as-is (beyond the step 0 comment) and we can see how it does.

PR #50, opened 2026-02-12
Deepen plugin-dev agent invocation for validation

withinfocus:
## 🎟️ Tracking

Discovered while testing new agent and plugin development.

## 📔 Objective

More explicitly invokes the two agents that Anthropic offers in their `plugin-dev` plugin for plugin validation and skill review.

### DIFF
--- a/.github/workflows/validate-plugins.yml
+++ b/.github/workflows/validate-plugins.yml
@@ -134,10 +134,7 @@ jobs:
         id: components
         if: steps.changed-files.outputs.has_components == 'true'
         continue-on-error: true
-        uses: anthropics/claude-code-action@1b8ee3b94104046d71fde52ec3557651ad8c0d71 # v1.0.29
-        env:
-          USE_AGENT_SDK: "true"
-          USE_SIMPLE_PROMPT: "true"
+        uses: anthropics/claude-code-action@f669191d7d1e67f08a54b0c11cf5683a9a391951 # v1.0.49
         with:
           anthropic_api_key: ${{ steps.get-kv-secrets.outputs.ANTHROPIC-CODE-REVIEW-API-KEY }}
           track_progress: true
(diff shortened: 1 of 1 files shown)


.github/workflows/validate-plugins.yml
```diff
@@ -134,10 +134,7 @@ jobs:
...
         if: steps.changed-files.outputs.has_components == 'true'
         continue-on-error: true
-        uses: anthropics/claude-code-action@1b8ee3b94104046d71fde52ec3557651ad8c0d71 # v1.0.29
-        env:
-          USE_AGENT_SDK: "true"
-          USE_SIMPLE_PROMPT: "true"
+        uses: anthropics/claude-code-action@f669191d7d1e67f08a54b0c11cf5683a9a391951 # v1.0.49
         with:
           anthropic_api_key: ${{ steps.get-kv-secrets.outputs.ANTHROPIC-CODE-REVIEW-API-KEY }}
           track_progress: true
```
theMickster, line 140: 🎨 While we are working in this workflow, will you please consider removing `track_progress: true` and `use_sticky_comment: true`?  The problem that we are facing is competing Claude reviews when the `ai-review` label is added.   The sticky comment gets overwritten several times making it hard to understand what's going on. Thanks!

theMickster: I am also okay if you want that in a follow-up PR.

withinfocus: I'll wait on that, since I see it in the other places we use this Action.

theMickster: @withinfocus 
Agreed that we are using it in both places, but I believe we are using it incorrectly in this repo because we are overwriting reviews. _[we are also using it incorrectly in the review workflow, but I need to organize that research before I present it]_

For example, [this Claude comment](https://github.com/bitwarden/ai-plugins/pull/46#issuecomment-3880029548) shows that two independent Claude agents overwrote one another because we're using the sticky & tracking parameters. 

The [Validate plugin structure and requirements summary](https://github.com/bitwarden/ai-plugins/actions/runs/21877855434) was first and left its comment. 
And then, [Review / Review summary](https://github.com/bitwarden/ai-plugins/actions/runs/21877855562#summary-63151745457) came along second and overwrote it. 

Therefore, I didn't even see the plugin validation comment.

withinfocus: https://github.com/bitwarden/ai-plugins/pull/51

PR #72, opened 2026-03-27
Minor changes to the first instructions for the agent and a summary comment header

theMickster:
## 🎟️ Tracking

[PM-34217](https://bitwarden.atlassian.net/browse/PM-34217)

## 📔 Objective

1. Step 1 was a thinking block with no data to think about — the agent hadn't fetched the diff yet, so it was reasoning in a vacuum before doing any actual work.
2. Commenting constraints lived in Step 1 but weren't actionable until Step 5 — rules about duplicate prevention and respecting human decisions now sit where comments are actually posted.
3. Two steps doing one job — "Read Existing Context" and "Understand the Change" were artificially split; the agent now gathers context and assesses the change in one pass with real data.

[PM-34217]: https://bitwarden.atlassian.net/browse/PM-34217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/plugins/bitwarden-code-review/skills/avoiding-false-positives/SKILL.md
+++ b/plugins/bitwarden-code-review/skills/avoiding-false-positives/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: avoiding-false-positives
-description: Use this skill to validate findings during Step 4 (Validate Findings). For each finding, run the rejection criteria and verification checks. If a finding fails any check, drop it.
+description: Use this skill to validate findings during a code review. For each finding, run the rejection criteria and verification checks. If a finding fails any check, drop it.
 ---
 
 # Validating Findings


plugins/bitwarden-code-review/skills/avoiding-false-positives/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: avoiding-false-positives
-description: Use this skill to validate findings during Step 5 (Validate Findings). For each finding, run the rejection criteria and verification checks. If a finding fails any check, drop it.
+description: Use this skill to validate findings during Step 4 (Validate Findings). For each finding, run the rejection criteria and verification checks. If a finding fails any check, drop it.
```
SaintPatrck: ⛏️ 🎨 
>Use this skill to validate findings during Step 4 (Validate Findings).

Could we rephrase this to be a little more generic instead of referencing an ambiguous "Step 4"? Two reasons:
- I could see this skill being useful to validate another reviewer's (or `claudebot`'s) comments outside of the code review agent's process. 
- If another agent or skill references a "Step 4", we don't want it to inadvertently invoke this skill.

theMickster: Yep, no problem

theMickster: Done

PR #80, opened 2026-04-08
Clean up security architecture skill's crypto sections

quexten:
## 📔 Objective

I did an initial review of the security architecture review skill. 

> - Never use MD5, SHA-1, or SHA-256 alone for password hashing (too fast, no salt by default)

This phrasing implies that MD5 or SHA-1 would be acceptable as part of a construction that has a work factor and salt, such as PBKDF2, but that is not true. SHA-1 and MD5 are no longer cryptographically secure hash functions, and should NEVER be used.

> - Salt must be unique per credential, cryptographically random, at least 128 bits

This is not true, salts do not need to have entropy, they *should* just be unique to ensure rainbow tables cannot be built / domain separation for brute-force attacks. But this does not mean that there needs to be cryptographic entropy. Email for instance is just as good from a security perspective.

> - Encryption keys must be stored separately from encrypted data (never in the same database)

This is not true. It is acceptable to have the keys stored in the same database, and we commonly do it. The CEK for vault items (cipher key) and the KEK/DEK for users (user key) are all stored in the same database as the vault items. This is fine, they are themselves encrypted.


abergs (dismissed):
lgtm

plugins/bitwarden-security-engineer/skills/reviewing-security-architecture/SKILL.md
```diff
@@ -74,7 +73,6 @@ public async Task<Cipher> GetCipher(Guid cipherId) {
 ### Encryption at Rest
 
 - All sensitive data must be encrypted at rest using AES-256 or equivalent
-- Encryption keys must be stored separately from encrypted data (never in the same database)
```
withinfocus, line 77: ⚠️ The intention here was to imply direct decrypt-ability from one to the other which I get is not the case for us, but per your comment this can just be elaborated.

plugins/bitwarden-security-engineer/skills/reviewing-security-architecture/SKILL.md
```diff
@@ -27,10 +27,9 @@ Review these aspects of token-based authentication:
 
 ### Credential Storage
 
-- Passwords must be hashed with a modern KDF: Argon2id (preferred), bcrypt, or PBKDF2 with high iteration count
-- Never use MD5, SHA-1, or SHA-256 alone for password hashing (too fast, no salt by default)
-- Salt must be unique per credential, cryptographically random, at least 128 bits
-- Consider pepper (application-level secret added to hash input) for defense in depth
+- Passwords must be hashed with a modern KDF: Argon2id (preferred), bcrypt, or PBKDF2 with high work factor and a unique salt
+- Never use raw cryptographic hash functions alone for password hashing (too fast, no salt by default)
+- Salt must be unique per credential
```
withinfocus: ❌ I would add "should" language instead; don't remove it though.

withinfocus (changes requested):
I'd like to see this go in the other direction where you're enhancing the rationale, not eliminating details. AI works well with specifics, and it's especially reactive when told what _not_ to do.

plugins/bitwarden-security-engineer/skills/reviewing-security-architecture/SKILL.md
quexten: How about:
```
- Cryptographic keys MUST be wrapped by other keys when stored in a database
```
or even
```
- Cryptographic keys MUST NEVER be stored directly accessible in a database, without being wrapped by another key
```

Besides that:
@withinfocus I've been thinking about making a separate cryptography skill to make the security architecture skill not grow too large. Do you think agree with the split, or should this be one large skill?

plugins/bitwarden-security-engineer/skills/reviewing-security-architecture/SKILL.md
quexten: Could you clarify where a pepper should be used within the context of Bitwarden? I do not know that we have a single use-case where it would be warranted.

plugins/bitwarden-security-engineer/skills/reviewing-security-architecture/SKILL.md
withinfocus: I like the latter statement.

Yeah, that sounds like a good idea. This skill isn't that large to me, but crypto could make it that way. Just ensure your crypto skill is about applications and not just definitions.

plugins/bitwarden-security-engineer/skills/reviewing-security-architecture/SKILL.md
withinfocus: I was referring to your statement of "salts do not need to have entropy, they should just be unique to ensure rainbow tables cannot be built / domain separation for brute-force attacks" as the language to include.

theMickster:
Curious what Claude Code has to say about our re-wording decisions... 🤔

plugins/bitwarden-security-engineer/skills/reviewing-security-architecture/SKILL.md
quexten: Okay, I think the new language should be better.

plugins/bitwarden-security-engineer/skills/reviewing-security-architecture/SKILL.md
quexten: Applied the latter.

theMickster (approved):

plugins/bitwarden-security-engineer/.claude-plugin/plugin.json
```diff
@@ -15,5 +15,7 @@
     "threat-modeling",
     "security-review"
   ],
-  "agents": ["./agents/bitwarden-security-engineer.md"]
+  "agents": [
```
withinfocus, line 18: ℹ️ `npm ci` is needed. There's a commit hook to use Prettier that will maintain these sorts of things.

quexten: Will make a follow-up PR, apologies. We should add npm ci as a blocking CI action as a second layer probably.

PR #88, opened 2026-04-16
PM-35200 - Enhance contributing doc to help be a better guide to teammates

theMickster:
## 🎟️ Tracking

[PM-35200](https://bitwarden.atlassian.net/browse/PM-35200)

## 📔 Objective

Our contributing doc needed to be less robotic/mechanical,  more inviting, and include a dedicate section about where ideas belong. As we progress, perhaps we trim the section, but for now, it's clear after speaking with engineers that the guidance is needed. 

[PM-35200]: https://bitwarden.atlassian.net/browse/PM-35200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ Example: `bitwarden-security-engineer`
 
 ### Tool Integration Plugins
 
-These connect Claude Code to external services the team already uses. They wrap APIs via MCP servers so Claude can read from and interact with those tools. They answer the question: _"I want Claude to talk to a service we use."_
+These connect Claude Code to external services the team already uses, so Claude can read from and act on those tools. They answer the question: _"I want Claude to securely integrate to a service we use."_
 
 If you find yourself context-switching between Claude Code and another tool to copy information back and forth, a tool integration plugin can bridge that gap.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,18 +1,22 @@
 # Contributing to Bitwarden AI Plugins
 
-Every engineer who has solved a recurring problem, documented a pattern, or automated a workflow has something worth sharing. This marketplace is where those ideas become plugins that the whole team can use. This guide helps you figure out where your idea fits, then walks through the mechanics of building and submitting it.
+Every engineer who has solved a recurring problem, documented a pattern, or automated a workflow has something worth sharing. This marketplace is where those ideas become plugins that the whole team can use. This guide helps you figure out where your work fits, then walks through the mechanics of building and submitting it.
 
 For general Bitwarden contribution practices, see our [Contributing Guidelines](https://contributing.bitwarden.com/contributing/).
 
-## Where Does Your Idea Belong?
+## Where Does Your Claude Tooling Belong?
 
-Plugins in this marketplace fall into three families. Understanding which one your idea belongs to will point you in the right direction.
+Plugins in this marketplace fall into three families. Repo-specific patterns usually belong closer to the code — see each repo's `.claude/CONTRIBUTING.md` for that. If your work is cross-repo and fits one of the families below, you're in the right place. If you're still unsure after reading them, raise a draft PR and maintainers will help find the right home.
 
 ### Persona Plugins
 
 These encode how a specific engineering role works at Bitwarden — the conventions, review standards, and decision frameworks that generic AI doesn't know. They answer the question: _"How does a software engineer, security engineer, or DevOps engineer work **here**?"_
 
-A persona plugin captures institutional knowledge that would otherwise live in someone's head or scattered across wiki pages. If your team has patterns, rules, or workflows that are specific to your discipline at Bitwarden, that's a persona plugin.
+Personas map to the _work_, not the title — when you're designing a system you're doing architecture work, and the matching persona is for you. Most engineers will reach for more than one persona across a week because engineers wear many hats.
+
+A persona plugin captures institutional knowledge that would otherwise live in someone's head or scattered across wiki pages. Persona plugins must clear three bars: the knowledge is institutional, domain-specific, and role-defining.
+
+Example: `bitwarden-security-engineer`
 
 ### Tool Integration Plugins
 
@@ -20,15 +24,13 @@ These connect Claude Code to external services the team already uses. They wrap 
 
 If you find yourself context-switching between Claude Code and another tool to copy information back and forth, a tool integration plugin can bridge that gap.
 
+Example: `bitwarden-atlassian-tools`
+
 ### Utility Plugins
 
 These improve the Claude Code development experience itself — setup, configuration, workflow analysis. They help every engineer regardless of role or domain. They answer the question: _"How can working with Claude Code be better for everyone?"_
 
----
-
-If you've encoded knowledge about how your role works, that's a persona plugin. If you want Claude to reach an external service, that's a tool integration. If you've improved how Claude Code itself works, that's a utility. See the [README](README.md) for the full list of current plugins.
-
-Not sure where your idea fits? Open a discussion or draft PR and maintainers will help find the right home.
+Examples: `bitwarden-init`
 
 ## Plugin Structure
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,7 +49,7 @@ plugins/your-plugin-name/
 └── .mcp.json               (MCP servers - optional)
 ```
 
-For detailed guidance on building each component, see the [Plugin Reference](https://docs.claude.com/en/docs/claude-code/plugins-reference.md).
+For detailed guidance on building each component, see the [Plugin Reference](https://code.claude.com/docs/en/plugins-reference).
 
 ## Adding a New Plugin
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,34 @@
 # Contributing to Bitwarden AI Plugins
 
-This guide covers everything you need to create, modify, and submit plugins for the Bitwarden AI Plugin Marketplace.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,9 +1,35 @@
 # Contributing to Bitwarden AI Plugins
 
-This guide covers everything you need to create, modify, and submit plugins for the Bitwarden AI Plugin Marketplace.
+Every engineer who has solved a recurring problem, documented a pattern, or automated a workflow has something worth sharing. This marketplace is where those ideas become plugins that the whole team can use. This guide helps you figure out where your idea fits, then walks through the mechanics of building and submitting it.
 
 For general Bitwarden contribution practices, see our [Contributing Guidelines](https://contributing.bitwarden.com/contributing/).
 
+## Where Does Your Idea Belong?
+
+Plugins in this marketplace fall into three families. Understanding which one your idea belongs to will point you in the right direction.
```
trmartin4: ❓ Is it convention that these all live under a `/plugins` directory, or could we add folders for each of these types (e.g. `/plugins/personas`) for easier discovery of the different types?

CONTRIBUTING.md
```diff
@@ -1,9 +1,35 @@
...
+
+### Persona Plugins
+
+These encode how a specific engineering role works at Bitwarden — the conventions, review standards, and decision frameworks that generic AI doesn't know. They answer the question: _"How does a software engineer, security engineer, or DevOps engineer work **here**?"_
+
+A persona plugin captures institutional knowledge that would otherwise live in someone's head or scattered across wiki pages. If your team has patterns, rules, or workflows that are specific to your discipline at Bitwarden, that's a persona plugin.
```
trmartin4: ❓ Perhaps a sentence here to clarify whether the expectation is that teams should add team-specific personas built on the core ones, or add team-specific patterns to the personas already there, or if it's an anti-pattern to have very team-specific guidance at this level (i.e. it's a process smell and we should consolidate our practices before asking Claude to follow them).

trmartin4 (commented):
This is really helpful!  Thank you for writing it up.  I had a couple of questions.

CONTRIBUTING.md
theMickster: The convention that I am seeing in the industry (primarily Anthropic & Microsoft Copilot's patterns) is to not nest these plugins like you are suggesting.

CONTRIBUTING.md
theMickster: I'd push back on team-specific personas (Platform, DIRT, Autofill, etc.) in the plugins repo. They aren't generic enough, and that kind of team-specific context usually belongs closer to the code — in `CLAUDE.md` files within each repo.

Stepping back to the broader workflow: I (and others) feel that no engineering team's software development process should be unique enough to warrant its own Claude persona. Where we see that, it's a signal to address as part of our SDLC maturity work, not to encode into a persona.

The "what" a team does to implement a feature belongs next to the code. The new [cipher-type-planner skill](https://github.com/bitwarden/clients/blob/main/.claude/skills/cipher-type-planner/SKILL.md) is a good example — it lives in clients because that proximity is what makes Claude effective with it. I'd expect a similar skill in sdk-internal eventually, for the same reason.

Does that help clarify where I'm coming from?

CONTRIBUTING.md
MGibson1: This sentence also assumes that ideas belong in plugins, but help determining _that_ would be good too. Certainly not all improvements belong even in this repository

CONTRIBUTING.md
MGibson1: In that case, the term "team" is probably the trigger.  Vertical might work better

CONTRIBUTING.md
```diff
@@ -1,9 +1,35 @@
...
+
+### Utility Plugins
+
+These improve the Claude Code development experience itself — setup, configuration, workflow analysis. They help every engineer regardless of role or domain. They answer the question: _"How can working with Claude Code be better for everyone?"_
+
+---
+
+If you've encoded knowledge about how your role works, that's a persona plugin. If you want Claude to reach an external service, that's a tool integration. If you've improved how Claude Code itself works, that's a utility. See the [README](README.md) for the full list of current plugins.
+
+Not sure where your idea fits? Open a discussion or draft PR and maintainers will help find the right home.
```
MGibson1: This seems to be a TLDR; but I'm not sure the horizontal break above communicates that

I also prefer these to be summaries at the top

CONTRIBUTING.md
```diff
@@ -1,9 +1,35 @@
...
+
+### Tool Integration Plugins
+
+These connect Claude Code to external services the team already uses. They wrap APIs via MCP servers so Claude can read from and interact with those tools. They answer the question: _"I want Claude to talk to a service we use."_
+
+If you find yourself context-switching between Claude Code and another tool to copy information back and forth, a tool integration plugin can bridge that gap.
+
+### Utility Plugins
+
+These improve the Claude Code development experience itself — setup, configuration, workflow analysis. They help every engineer regardless of role or domain. They answer the question: _"How can working with Claude Code be better for everyone?"_
```
MGibson1, line 31: I think this could use an example which explains why the utility is needed and not appropriate for claude.md files.

CONTRIBUTING.md
```diff
@@ -1,9 +1,35 @@
 # Contributing to Bitwarden AI Plugins
 
-This guide covers everything you need to create, modify, and submit plugins for the Bitwarden AI Plugin Marketplace.
+Every engineer who has solved a recurring problem, documented a pattern, or automated a workflow has something worth sharing. This marketplace is where those ideas become plugins that the whole team can use. This guide helps you figure out where your idea fits, then walks through the mechanics of building and submitting it.
 
 For general Bitwarden contribution practices, see our [Contributing Guidelines](https://contributing.bitwarden.com/contributing/).
 
+## Where Does Your Idea Belong?
```
MGibson1: Not sure this title is appropriate. Per your response to @trmartin4, this implies you define the _type_ of skill/plugin via it's location. That doesn't appear to be accurate.

However, I also don't actually know how to specify that your file is each particular type of idea. I think each section here deserves a subsection describing the steps needed to produce that particular type. I'm sure Claude could help users figure it out, but the goal of these docs is to avoid having to swap tools

MGibson1 (changes requested):
A few of these deserve addressing to make sure this distinction is clear and actionable

CONTRIBUTING.md
theMickster: I believe that I have addressed the concerns in 9498b815cff68a337c5ee8011e5b557b92291066

CONTRIBUTING.md
theMickster: I believe that I have addressed the concerns in 9498b815cff68a337c5ee8011e5b557b92291066

CONTRIBUTING.md
theMickster: I believe that I have addressed the concerns in 9498b815cff68a337c5ee8011e5b557b92291066

CONTRIBUTING.md
theMickster: I believe that I have addressed the concerns in 9498b815cff68a337c5ee8011e5b557b92291066

CONTRIBUTING.md
theMickster: I believe that I have addressed the concerns in 9498b815cff68a337c5ee8011e5b557b92291066

theMickster:
Hi @trmartin4  and @MGibson1 👋🏼
Please take a peek at the second round of phrasing in the document and let me know how it reads. Thanks!

MGibson1 (dismissed):

CONTRIBUTING.md
Banrion: ❓Can we consider changing this to the [docs link](https://code.claude.com/docs/en/plugins-reference) instead of the link to the markdown file?

Banrion, line 55: ❓ From [this](https://github.com/bitwarden/ai-plugins/pull/75#discussion_r3027554854) comment on another pull request, it's clear we want teams to maintain plugins. Should we add a line to this section to call out when to add a team as a maintainer?

theMickster: Not yet, let's focus solely on getting the path to contributing work first. Then tackle ownership in follow-up PRs.

theMickster: Addressed in f0480cb509e27819c7b0d38b16b1077b7f7b27f0

Banrion (dismissed):

trmartin4 (dismissed):
Thank you for making the adjustments!

CONTRIBUTING.md
```diff
@@ -1,9 +1,37 @@
...
+
+Personas map to the _work_, not the title — when you're designing a system you're doing architecture work, and the matching persona is for you. Most engineers will reach for more than one persona across a week because engineers wear many hats.
+
+A persona plugin captures institutional knowledge that would otherwise live in someone's head or scattered across wiki pages. Persona plugins must clear three bars: the knowledge is institutional, domain-specific, and role-defining.
+
+Example: `bitwarden-security-engineer`
+
+### Tool Integration Plugins
+
+These connect Claude Code to external services the team already uses. They wrap APIs via MCP servers so Claude can read from and interact with those tools. They answer the question: _"I want Claude to talk to a service we use."_
```
dasien: I would be careful with this singular definition of a tool integration.  They don't always 'wrap APIs via MCP servers', for example.

theMickster: That's a good point.  I will work to loosen the wording.

theMickster: Refactored in ba8febcf5124b101e3e84bf55816942254140ba7

SaintPatrck (approved):

PR #105, opened 2026-04-27
Allow read-only gh and git in code-review-local command

SaintPatrck:
## 🎟️ Tracking

N/A — internal plugin polish.

## 📔 Objective

Bumps `bitwarden-code-review` to **1.9.1**.

The `/bitwarden-code-review:code-review-local` command currently has no `allowed-tools` frontmatter, so every invocation prompts for permission on the read-only `gh` and `git` calls it needs to fetch PR metadata, diff, and existing review threads. This PR mirrors what `code-review` already declares so those calls are pre-authorized for the duration of the command:

- Adds `allowed-tools` to the `code-review-local` command frontmatter covering the read-only `gh pr view`, `gh pr diff`, `gh pr checks`, `gh pr list`, `gh api graphql`, `git show`, `git log`, `git diff` patterns the command actually uses.
- Scopes `Write` to the two output files (`review-summary.md`, `review-inline-comments.md`) rather than granting it broadly — keeps the command focused on its documented output and prevents incidental writes elsewhere.
- No behavioral change to the review itself; this only removes per-invocation prompting friction.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -25,7 +25,7 @@
     {
       "name": "bitwarden-code-review",
       "source": "./plugins/bitwarden-code-review",
-      "version": "1.9.0",
+      "version": "1.9.1",
       "description": "Comprehensive code review system with organization-wide standards."
     },
     {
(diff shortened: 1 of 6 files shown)


theMickster (approved):

SaintPatrck:
Thanks @theMickster

PR #106, opened 2026-04-28
Add addressing-code-review-comments skill

SaintPatrck:
## 🎟️ Tracking

N/A — internal skill addition.

## 📔 Objective

Adds a new `addressing-code-review-comments` skill to the `bitwarden-code-review` plugin, scoped to local PR review comment work (a developer addressing comments on their own PR with Claude's help).

The skill emphasizes:
- Verifying each suggestion against the codebase before implementing
- Surfacing ambiguity to the user before touching code
- Presenting fixes or pushback drafts to the user without performative agreement
- Claude does **not** post on the user's behalf — only drafts plain text the user can paste

The plugin already has a `bitwarden-code-reviewer` agent for *producing* reviews. This new skill covers the inverse workflow: a developer who has *received* review comments and wants help working through them locally. The two have different triggering conditions and different rules of engagement, so they live as separate components.

Plugin bumped to **1.10.0** (MINOR — new skill), with a Keep-a-Changelog entry.

<details>
<summary>⏺ Skill validation: addressing-code-review-comments</summary>

```markdown

  Ran a controlled benchmark with 8 synthetic review comments designed to probe specific failure modes (YAGNI, wrong-but-confident claims, vague pressure, authority bias, non-equivalent rewrites, thread reversals, tradeoffs, reviewer-missing-context). Each case ran with the skill and without, across Opus 4.7, Sonnet 4.6, and Haiku — then a targeted edit to address one persistent failure mode, validated with reruns to confirm reproducibility.

  Cross-model results (original skill)

  ┌────────────┬────────────┬────────────┬─────────┬──────────────┐
  │   Model    │ Lift cases │ Harm cases │ Token Δ │ Wall-clock Δ │
  ├────────────┼────────────┼────────────┼─────────┼──────────────┤
  │ Opus 4.7   │      1 / 8 │          0 │   +1.8% │      neutral │
  ├────────────┼────────────┼────────────┼─────────┼──────────────┤
  │ Sonnet 4.6 │      2 / 8 │          0 │   −6.5% │        −9.2% │
  ├────────────┼────────────┼────────────┼─────────┼──────────────┤
  │ Haiku      │      1 / 8 │          0 │   +0.7% │         +14% │
  └────────────┴────────────┴────────────┴─────────┴──────────────┘

  Zero harm across all 24 trials. The skill compensates more on weaker models — Sonnet+skill ≈ Opus baseline on this benchmark.

  Targeted edit + validation (Haiku)

  Eval-7 (genuine tradeoff: "either could be defensible — what's the intent?") failed on all four prior conditions: Opus±skill, Sonnet±skill, Haiku-original-skill. Every model confidently picked a side instead of surfacing the tradeoff to the user.

  I added a ## When the Reviewer Asks a Tradeoff Question section with explicit guidance (don't pick a side, lay out both, ask the user) and a Bad/Good example pair. Re-ran on Haiku, plus 2 rerun trials of eval-5 to verify a suspected regression wasn't real.

  ┌──────────────────────────────────┬─────────┬─────────┬───────────┬────────┐
  │          Configuration           │ Correct │ Partial │ Incorrect │  Lift  │
  ├──────────────────────────────────┼─────────┼─────────┼───────────┼────────┤
  │ Haiku, original skill            │       5 │       2 │         1 │    1/8 │
  ├──────────────────────────────────┼─────────┼─────────┼───────────┼────────┤
  │ Haiku, edited skill (post-rerun) │       7 │       1 │         0 │ 3/8 ⭐ │
  └──────────────────────────────────┴─────────┴─────────┴───────────┴────────┘

  The edit moved the skill from marginal lift (1/8) to meaningful lift (3/8) on Haiku, with zero harm.

  Where the skill earns its keep

  Eval-8 — reviewer claims SDK function is non-nullable, says "remove the else branch." The branch is real defensive code (legacy AesCbc keys return null; sibling files use the same guard). Without the skill, models drifted toward "polite middle-ground" — Opus capitulated to deleting the branch outright, Sonnet proposed deleting it under both confirmation outcomes. With the skill, both held firm: Opus refused changes pending SDK signature, Sonnet pushed back with citations
  to SymmetricCryptoKey's legacy types and the parallel renderer-service guard. Headline lift across both Opus and Sonnet.

  Eval-3 — reviewer comments "this whole pattern is wrong" with no specifics. Sonnet without the skill invented a coherent interpretation (the migration duplicates state with RendererBiometricsService.enrollPersistent) and proposed concrete refactors. Plausible-looking, pure speculation. With the skill, Sonnet stayed in clarification mode and asked the reviewer to specify. Skill prevented confident-but-wrong over-interpretation.

  Eval-5 — reviewer proposes a non-equivalent "simplification." Haiku without the skill confidently signed off on a rewrite that would have introduced a real bug (different behavior when enrolledKeyId != null && currentKeyId == null). With the skill, Haiku enumerated the truth table, caught the divergent case, and pushed back with the counter-example. Strongest "safety" result in the benchmark.

  Eval-7 — genuine tradeoff with no objectively correct answer. Pre-edit: every model picked a side. Post-edit: Haiku produced exactly the intended behavior — laid out Option A vs Option B with pro/con, flagged its lean as a "read" not the answer, and asked the user to decide before drafting. Quote: "Decision needed from you (the author), not a technical verdict… Before you reply to the reviewer, what's your intent?"

  Where the skill saves cost

  On Sonnet, the skill saves tokens (−6.5%) and wall clock (−9.2%) on average. Counterintuitive but consistent: the skill's "verify and ask before acting" instruction short-circuits speculative exploration paths. In eval-3 alone, Sonnet without the skill burned 47.7k tokens building hypothetical refactors; with the skill, 38.0k tokens reaching a cleaner clarification response. Same shape on eval-8.

  The edit on Haiku adds ~5% tokens (longer skill text) but cuts wall clock 34% — explicit guidance prevents speculative paths.

  Bottom line

  - 24 trials × 3 models, 0 harm cases
  - Lift on every model tier, growing as model capability decreases (1 / 2 / 3 cases on Opus / Sonnet / Haiku-edited)
  - Net cost negative on Sonnet and net cost neutral elsewhere (saves wall clock on Haiku, marginal token overhead on Opus)
  - One known failure mode (eval-7) targeted, fixed, and rerun-confirmed
  - Edit applied to SKILL.md — 25 lines added, voice-matched to existing skill style
```

</details>

<details>
<summary>⏺ Sample responses showing skill behavior</summary>

```markdown
  Below are excerpts from the actual benchmark runs — synthetic reviewer comments paired with how the model responded with vs. without the skill. All quotes are verbatim from iteration-{1..4}/eval-N/response.md.

  ---
  Eval-7 — Tradeoff with no objectively correct answer (Haiku, post-edit)

  Reviewer: "What happens if enrollPersistent succeeds but setBiometricProtectedUnlockKeyForUser fails? We end up in a half-migrated state where the persistent key is enrolled but the unlock binding is stale. Should we wrap these in a transaction-like rollback, or is the current behavior acceptable since needsMigration() will detect the inconsistency on the next run? Either could be defensible — what's the intent?"

  Without the edit (model picks a side):

  ▎ "Choose: CLARIFICATION + minor documentation improvement. Self-healing is actually correct here… Add a comment explaining the failure tolerance design."

  With the edit (model surfaces the tradeoff):

  ▎ "Decision needed from you (the author), not a technical verdict. This is a genuine tradeoff:
  ▎
  ▎ Option A: Current approach (rely on retry detection) — Pro: simpler code, no rollback machinery; Con: brief window where unlock binding is missing while persistent key exists.
  ▎
  ▎ Option B: Add explicit rollback or transaction wrapper — Pro: guarantees atomicity; Con: adds complexity (what if the rollback itself fails?).
  ▎
  ▎ My read: The current design is defensible if needsMigration() reliably detects the half-migrated state… However, if you're concerned about the user experience during that window, explicit rollback is the safer choice.
  ▎
  ▎ Before you reply to the reviewer, what's your intent? Are you comfortable with the retry-based recovery, or do you want to strengthen this with rollback logic?"
  ▎
  ▎ Draft reply: "Pending your decision. Once you confirm which direction you want to take, I'll write a reply that either: (1) explains why retry-based is acceptable, or (2) acknowledges the gap and commits to rollback logic."

  This was the only case that failed across all four prior conditions (Opus±skill, Sonnet±skill, Haiku-original-skill). The targeted edit fixed it.

  ---
  Eval-8 — Reviewer makes confident-but-unverifiable claim (Opus)

  Reviewer: "The else branch is dead code — get_key_id_for_symmetric_key always returns a non-null value for any valid SymmetricKey input. Remove the null check and the else branch entirely; this just adds noise."

  Without the skill (model offers a "polite middle-ground" that gives away defensive code):

  ▎ "Recommendation: Fix — partial agreement. Remove the else branch and its setBiometricEnrolledKeyId(userId, null) call, matching the existing pattern in RendererBiometricsService.enrollPersistent. Keep the if (keyId != null) guard.
  ▎
  ▎ Justification: I will not fully accept the reviewer's framing… But the reviewer is correct that the else branch adds noise."

  With the skill (model holds firm pending verification):

  ▎ "Recommendation: Clarification. Don't change code yet. The reviewer is making a contract claim about an SDK function whose type definition isn't even present in the currently installed SDK package, and the same file's needsMigration is written assuming the opposite…
  ▎
  ▎ Implementation notes: No code changes yet — waiting on SDK contract clarification."

  The without-skill response would have shipped a removal of defensive code based on an unverified claim. The with-skill response correctly refused.

  ---
  Eval-5 — Non-equivalent "simplification" hides a bug (Haiku)

  Reviewer: "This branching is overcomplicated. Simplify with: return enrolledB64 !== currentB64 ? 'needsMigration' : 'noMigrationNeeded'; One conditional, no flags."

  Without the skill (model signs off on a rewrite that introduces a bug):

  ▎ "Result: The two implementations are equivalent. The reviewer's approach eliminates boolean flags and reduces nesting depth.
  ▎
  ▎ Recommendation: Fix — Accept the simplification. Justification: Logically equivalent — both handle all cases identically. Improves readability… The reviewer's version is the stronger approach.
  ▎
  ▎ Draft reply: 'You're right—this is cleaner.'"

  With the skill (model catches the divergent case):

  ▎ "The divergence: When enrolledKeyId is a non-null B64 string and currentKeyId is null, the original code correctly returns noMigrationNeeded because the condition isV1ToV2Migration (requires currentKeyId != null) fails. The proposed rewrite would instead compare "B64str" !== null, which is true, and incorrectly return needsMigration.
  ▎
  ▎ Recommendation: Pushback. The proposed simplification changes behavior in a case the original correctly handles… The reviewer may not be aware that enrolledKeyId can be a non-null string while currentKeyId is null, or that this condition should explicitly NOT trigger migration."

  The without-skill response would have shipped a behavioral regression. The with-skill response caught it via truth-table enumeration. Strongest safety result in the benchmark.

  ---
  Eval-3 — Vague pressure with no specifics (Sonnet)

  Reviewer (the entire comment): "This whole pattern is wrong."

  Without the skill (model invents a coherent interpretation and proposes concrete code changes):

  ▎ "After reading it alongside the other changes in the PR, I think I understand the concern. RendererBiometricsService.enrollPersistent was updated in this PR to handle the enrolled-key-ID tracking itself… But BiometricPersistentMigration.runMigrations then calls enrollPersistent and re-derives and re-sets the key ID a second time…
  ▎
  ▎ Recommendation: Fix. My proposed fix:
  ▎ - Remove the key-ID re-derivation block (lines 70–80)
  ▎ - Remove the setBiometricProtectedUnlockKeyForUser call from runMigrations
  ▎ - Remove SdkService from the migration's constructor
  ▎ - Address the fragile coupling in MainBiometricsService.enrollPersistent"

  With the skill (model refuses to act on guesses):

  ▎ "The comment 'This whole pattern is wrong' is anchored at line 18 — the class declaration. It is vague but points at the class as a whole, not a single line. To interpret it correctly, the code must be read in context… None of [the four candidate concerns I identified] constitutes 'the whole pattern is wrong.'
  ▎
  ▎ Recommendation: Clarification. Justification: the comment is non-actionable as written. Without the reviewer naming what is wrong, any rewrite is guesswork. The respectful, technically rigorous move is to ask the reviewer to specify the concern.
  ▎
  ▎ Implementation notes: No code changes pending reviewer clarification."

  The without-skill response would have proposed a substantial refactor based on an invented interpretation. The with-skill response correctly stayed in clarification mode.

  ---
  Eval-2 — Wrong-but-confident technical claim (Sonnet, with skill)

  Reviewer: "JavaScript's !== does reference equality on strings — this comparison won't work as expected when the strings come from different sources. Use String.prototype.localeCompare() or a deep-equal helper instead."

  Model response (with skill):

  ▎ "This claim is technically incorrect and conflates JavaScript's string semantics with those of reference-type languages such as Java or C#…
  ▎
  ▎ In JavaScript and TypeScript:
  ▎ - Strings are primitive values, not objects. === and !== perform value comparison on primitives, per the ECMAScript Strict Equality Comparison algorithm (7.2.16): for two operands of type String, the result is true iff they are the same sequence of code units…
  ▎ - localeCompare would actually be a worse fit here: it performs locale-aware collation (which can treat different code-point sequences as equivalent under Unicode normalization or locale rules), and it returns a number so it must be compared against 0…
  ▎
  ▎ The existing test 'should return noMigrationNeeded when enrolled key ID matches current key ID' empirically confirms !== returns false when two independently-produced base64 strings have the same contents — this is exactly the scenario the reviewer claims is broken."

  Both Opus and Sonnet handled this case correctly with or without the skill — but it's a useful counterexample showing the model isn't just sycophantically refusing every comment.
```

</details>

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -25,7 +25,7 @@
     {
       "name": "bitwarden-code-review",
       "source": "./plugins/bitwarden-code-review",
-      "version": "1.9.1",
+      "version": "1.10.0",
       "description": "Comprehensive code review system with organization-wide standards."
     },
     {
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A curated collection of plugins for AI-assisted development at Bitwarden. Enable
 | ------------------------------------------------------------------- | ------- | ------------------------------------------------------------------------------------------------------------------- |
 | [bitwarden-tech-lead](plugins/bitwarden-tech-lead/)                 | 2.0.0 | Software architect for technical planning, architecture reviews, and implementation phasing                         |
 | [bitwarden-atlassian-tools](plugins/bitwarden-atlassian-tools/)     | 2.2.3   | Read-only Atlassian access via MCP server with deep Jira issue research skill                                       |
-| [bitwarden-code-review](plugins/bitwarden-code-review/)             | 1.9.1   | Autonomous code review agent following Bitwarden engineering standards with GitHub integration                      |
+| [bitwarden-code-review](plugins/bitwarden-code-review/)             | 1.10.0  | Autonomous code review agent following Bitwarden engineering standards with GitHub integration                      |
 | [bitwarden-delivery-tools](plugins/bitwarden-delivery-tools/)       | 1.0.0   | Generic delivery workflow skills for committing, PR creation, preflight checks, and change labeling                 |
 | [bitwarden-devops-engineer](plugins/bitwarden-devops-engineer/)     | 0.1.1   | DevOps engineering assistant: workflow compliance linting, action security auditing, and org-wide CI/CD remediation |
 | [bitwarden-init](plugins/bitwarden-init/)                           | 1.1.0   | Initialize and enhance CLAUDE.md files with Bitwarden's standardized template format                                |
--- a/plugins/bitwarden-code-review/.claude-plugin/plugin.json
+++ b/plugins/bitwarden-code-review/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "bitwarden-code-review",
-  "version": "1.9.1",
+  "version": "1.10.0",
   "description": "Comprehensive code review system with organization-wide standards.",
   "author": {
     "name": "Bitwarden",
--- a/plugins/bitwarden-code-review/CHANGELOG.md
+++ b/plugins/bitwarden-code-review/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to the Bitwarden Code Review Plugin will be documented in th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.10.0] - 2026-04-28
+
+### Added
+
+- New `addressing-code-review-comments` skill that guides Claude when a developer is working through PR review comments locally — emphasizes verifying each suggestion against the codebase, surfacing ambiguity before touching code, and presenting fixes or pushback drafts to the user without performative agreement
+
 ## [1.9.1] - 2026-04-27
 
 ### Changed
--- a/plugins/bitwarden-code-review/README.md
+++ b/plugins/bitwarden-code-review/README.md
@@ -15,20 +15,22 @@ This plugin provides an autonomous code review agent that conducts thorough, pro
 - **Structured Thinking**: Uses explicit reasoning blocks to improve review quality and consistency
 - **Confidence Scoring**: Pre-filters findings with a 0-100 confidence score (≥75 threshold) before validation to reduce false positives
 
+## Skills
+
+| Skill                                                                             | Triggers                                            | Purpose                                                                                                              |
+| --------------------------------------------------------------------------------- | --------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------- |
+| [`classifying-review-findings`](./skills/classifying-review-findings/SKILL.md)         | "classify finding", "severity"                      | 5-tier severity system (CRITICAL / IMPORTANT / DEBT / SUGGESTED / QUESTION) with emoji and label mapping             |
+| [`avoiding-false-positives`](./skills/avoiding-false-positives/SKILL.md)               | "validate finding", "verify before posting"         | Rejection criteria and verification checks that drop low-confidence findings before they reach a comment             |
+| [`posting-bitwarden-review-comments`](./skills/posting-bitwarden-review-comments/SKILL.md) | "post inline comment", "post PR comment"            | Inline PR comment formatting per Bitwarden standards (severity emojis, explanation, actionable suggestion)           |
+| [`posting-review-summary`](./skills/posting-review-summary/SKILL.md)                   | "post summary", "summary comment"                   | Final summary comment handling — routes to sticky comment, GitHub Actions MCP tool, or local file based on context   |
+| [`reviewing-dependency-changes`](./skills/reviewing-dependency-changes/SKILL.md)       | "package.json", "Renovate PR", "dependency manifest" | Flags dependency manifest changes for AppSec approval, version-bump significance, and lock-file hygiene              |
+| [`addressing-code-review-comments`](./skills/addressing-code-review-comments/SKILL.md) | "address review comments", "respond to PR feedback" | Guides developers working through review comments locally — verify before implementing, surface ambiguity, no performative agreement |
+
 ## Architecture
 
 ### Code Review Agent
 
 The plugin provides a single agent (`bitwarden-code-reviewer`) that follows a linear 7-step review process — from context gathering through validation to posting. See [`AGENT.md`](./agents/bitwarden-code-reviewer/AGENT.md) for the full flow.
-
-### Skills
-
-The agent leverages specialized skills:
-
-- **`classifying-review-findings`**: Determines severity levels and validates finding criteria
-- **`posting-bitwarden-review-comments`**: Formats inline PR comments following Bitwarden standards
-- **`posting-review-summary`**: Posts or updates summary comments (handles sticky comment vs local file)
-- **`avoiding-false-positives`**: Validates findings against framework patterns and conventions
 
 ### Finding Classification
 
@@ -48,15 +50,7 @@ bitwarden-code-review/
 ├── commands/
 │   ├── code-review/                          # Code review command
 │   └── code-review-local/                    # Local review command
-├── skills/
-│   ├── avoiding-false-positives/
-│   │   └── SKILL.md                          # False positive prevention
-│   ├── classifying-review-findings/
-│   │   └── SKILL.md                          # Severity classification
-│   ├── posting-bitwarden-review-comments/
-│   │   └── SKILL.md                          # Inline comment formatting
-│   └── posting-review-summary/
-│       └── SKILL.md                          # Summary comment handling
+├── skills/                                   # See Skills table above
 ├── tests/
 │   └── TESTING.md                            # Test plan and validation
 └── README.md                                 # This file
--- a/plugins/bitwarden-code-review/agents/bitwarden-code-reviewer/AGENT.md
+++ b/plugins/bitwarden-code-review/agents/bitwarden-code-reviewer/AGENT.md
@@ -1,6 +1,6 @@
 ---
 name: bitwarden-code-reviewer
-version: 1.9.1
+version: 1.10.0
 description: Conducts thorough code reviews following Bitwarden standards. Finds all issues first pass, avoids false positives, respects codebase conventions. Invoke when user mentions "code review", "review code", "review", "PR", or "pull request".
 model: opus
 skills: avoiding-false-positives, classifying-review-findings, posting-bitwarden-review-comments, posting-review-summary, reviewing-dependency-changes
--- a/plugins/bitwarden-code-review/skills/addressing-code-review-comments/SKILL.md
+++ b/plugins/bitwarden-code-review/skills/addressing-code-review-comments/SKILL.md
@@ -44,6 +44,31 @@ Lead with technical reasoning, reference the specific code or constraint, ask a 
 
 If you recommended pushback and then find the reviewer was right, say so plainly and move on. Skip the apology.
 
+## When the Reviewer Asks a Tradeoff Question
+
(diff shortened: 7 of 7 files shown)


theMickster:
💭 When you have a chance, I am curious about the output from the skill. Do you have some cycles to toss a few prompts and responses in a expandable `<details>` block in either the description or a comment?

theMickster:
Hi @SaintPatrck 
A local multi agent code review brought up the following finding in the plugin that I think it's worth a quick cleanup. 
We have tended to combine these two listings into one new `## Skills` table in the more recently added plugins; so, perhaps, that's the best path forward. 🤔

<details> 
<summary> ♻️  Plugin README skill inventory is now two skills behind reality </summary>

`plugins/bitwarden-code-review/README.md:26-33` and `plugins/bitwarden-code-review/README.md:51-63`

The plugin README maintains two explicit skill inventories — a prose "Skills" section (lines 26-33) and a "Directory Structure" tree (lines 51-63). Both currently list only 4 skills:

  - `classifying-review-findings`
  - `posting-bitwarden-review-comments`
  - `posting-review-summary`
  - `avoiding-false-positives`

  Missing from both:
  - `reviewing-dependency-changes` (added in v1.9.0 — pre-existing drift)
  - `addressing-code-review-comments` (added in this PR)

Two of six skills (~33%) are absent from the README's own inventory. The PR that adds a new skill is the natural place to refresh the inventory. Suggested fix: add both skills to the prose list and the directory tree as part of this PR (or a small follow-up).

  ```markdown
  ### Skills

  The agent leverages specialized skills:

  - **`classifying-review-findings`**: Determines severity levels and validates finding criteria
  - **`posting-bitwarden-review-comments`**: Formats inline PR comments following Bitwarden standards
  - **`posting-review-summary`**: Posts or updates summary comments (handles sticky comment vs local file)
  - **`avoiding-false-positives`**: Validates findings against framework patterns and conventions
  - **`reviewing-dependency-changes`**: Flags dependency manifest changes for AppSec approval, version-bump significance, lock-file hygiene
  - **`addressing-code-review-comments`**: Guides developers working through review comments locally — verify before implementing, surface ambiguity, no performative agreement
  ```
</details>

SaintPatrck:
> Hi @SaintPatrck 
> A local multi agent code review brought up the following finding in the plugin that I think it's worth a quick cleanup. 
> We have tended to combine these two listings into one new `## Skills` table in the more recently added plugins; so, perhaps, that's the best path forward. 🤔
> 
> <details> 
> <summary> ♻️  Plugin README skill inventory is now two skills behind reality </summary>
> 
> `plugins/bitwarden-code-review/README.md:26-33` and `plugins/bitwarden-code-review/README.md:51-63`
> 
> The plugin README maintains two explicit skill inventories — a prose "Skills" section (lines 26-33) and a "Directory Structure" tree (lines 51-63). Both currently list only 4 skills:
> 
>   - `classifying-review-findings`
>   - `posting-bitwarden-review-comments`
>   - `posting-review-summary`
>   - `avoiding-false-positives`
> 
>   Missing from both:
>   - `reviewing-dependency-changes` (added in v1.9.0 — pre-existing drift)
>   - `addressing-code-review-comments` (added in this PR)
> 
> Two of six skills (~33%) are absent from the README's own inventory. The PR that adds a new skill is the natural place to refresh the inventory. Suggested fix: add both skills to the prose list and the directory tree as part of this PR (or a small follow-up).
> 
>   ```markdown
>   ### Skills
> 
>   The agent leverages specialized skills:
> 
>   - **`classifying-review-findings`**: Determines severity levels and validates finding criteria
>   - **`posting-bitwarden-review-comments`**: Formats inline PR comments following Bitwarden standards
>   - **`posting-review-summary`**: Posts or updates summary comments (handles sticky comment vs local file)
>   - **`avoiding-false-positives`**: Validates findings against framework patterns and conventions
>   - **`reviewing-dependency-changes`**: Flags dependency manifest changes for AppSec approval, version-bump significance, lock-file hygiene
>   - **`addressing-code-review-comments`**: Guides developers working through review comments locally — verify before implementing, surface ambiguity, no performative agreement
>   ```
> </details>

@claude fix this

theMickster:
Looks like the last commit threw something off with the validation/linting as those CI Actions appear to be hung.
The last commit bumped the version of the plugin to 1.10.1 but I don't think we want that. 
Also, the commit wasn't signed with make me a bit uneasy 😬

theMickster (approved):

SaintPatrck:
>Looks like the last commit threw something off

Ha. Yea. Guess we need to make some tweaks so Claude doesn't do that.

PR #119, opened 2026-05-14
bitwarden-init: add user/workspace CLAUDE.md scopes

withinfocus:
## 🎟️ Tracking

Inspired by [Claude-ing with Mick](https://bitwarden.atlassian.net/wiki/spaces/~712020eaf9a0d63a034f1ab7de1befe0802944/pages/2860417095/Claude-ing+with+Mick) — Mick's wiki page describes a three-level `CLAUDE.md` hierarchy (`~/.claude/`, `<bw-source-root>/.claude/`, `<repo>/`) that Claude Code automatically concatenates by walking up the directory tree. The plugin previously only handled the repo level.

## 📔 Objective

Extend `bitwarden-init` so engineers can set up the upper two scopes through a guided Q&A, without imposing one engineer's wording as the Bitwarden standard.

### New commands

| Command | Target |
| --- | --- |
| `/bitwarden-init:init-user` | `~/.claude/CLAUDE.md` (universal behavioral rules) |
| `/bitwarden-init:init-workspace` | `<bw-source-root>/.claude/CLAUDE.md` (Bitwarden domain context applied to every cloned repo) |
| `/bitwarden-init:enhance-user` | Refresh user-level by appending missing modules |
| `/bitwarden-init:enhance-workspace` | Refresh workspace-level by appending missing modules |

The existing `/bitwarden-init:init` and `/bitwarden-init:enhance` (repo scope, 11-section project template) are unchanged.

### Q&A flow

Every command uses `AskUserQuestion` for every decision — no free-text parsing:

1. **Resolve target.** `init-workspace` auto-detects the workspace root by walking up from `cwd` looking for an ancestor with 2+ sibling git repositories; always asks for confirmation.
2. **Detect existing file.** Offers Merge / Replace / Abort. `enhance-*` commands refuse if the file is missing and point at the matching `init-*`.
3. **Module selection.** Multi-select from the template library.
4. **Render preview.** Concatenates selected modules under a generated header that notes the version and the “one engineer's opinion, edit freely” framing.
5. **Diff + confirm.** Shows `diff -u` against the existing file (or the full rendered content if creating fresh), then asks `Apply` / `Show diff again` / `Cancel`.
6. **Backup, then write.** Copies the existing file to `<target>.bak-<ISO-timestamp>` before any overwrite. Surfaces the backup path in the success message.

### Module library

10 user-level + 8 workspace-level markdown stubs under `plugins/bitwarden-init/templates/`. Each module is 5–25 lines, starts with its own `##` heading so concatenation produces well-formed Markdown, and uses `[YOUR-PREFERENCE]` placeholders where opinion is sharp (e.g. the “no comments by default” rule) so users can invert without rewriting from scratch.

| Scope | Modules |
| --- | --- |
| User | intellectual-honesty, think-before-coding, simplicity, no-comments-default, surgical-changes, goal-driven-execution, verify-references, security-baseline, testing-discipline, code-review-lockdown |
| Workspace | overview, zero-knowledge-invariant, cryptography-choices, security-principles (P01–P06), core-vocabulary, agent-planning-requirements, operating-defaults, sources-of-truth |

### Design notes

- **Scope-explicit commands, not a single `--scope=` flag.** Scope is too consequential to bury in an argument; distinct slash entries are discoverable and let each have tailored prompts.
- **Modules, not Mick's template verbatim.** The wiki page is explicit that the content is one user's opinion. Shipping it as a checked-in standard would re-impose it on the marketplace.
- **No interaction with the existing `/enhance`.** The 11-section project template is wrong for behavioral and domain files — separate code path.

### Version + changelog

- Bumped 1.1.0 → 1.2.0 via `./scripts/bump-plugin-version.sh bitwarden-init 1.2.0`.
- CHANGELOG entry under `### Added` + `### Changed`.

### Validation

- `npm run lint` — clean (prettier + cspell; added `lockdown`, `HMAC`, `worktree`, `worktrees`, `docstrings` to `.cspell.json`)
- `./scripts/validate-plugin-structure.sh bitwarden-init` — 0 errors, 0 warnings (pre-existing "README should include usage examples" warning was fixed in this PR)
- `./scripts/validate-marketplace.sh bitwarden-init` — 0 errors
- `plugin-dev:plugin-validator` agent — clean

## ✅ Test plan

- [ ] Install the marketplace locally and verify the four new commands appear in the `/bitwarden-init:` slash menu
- [ ] Run `/bitwarden-init:init-user` against a non-existent `~/.claude/CLAUDE.md` — pick a couple of modules, confirm Apply, verify the file is created with the expected headings and the generated header banner
- [ ] Re-run `/bitwarden-init:init-user` against the just-written file — pick **Replace**, verify a `.bak-<timestamp>` is written next to the original
- [ ] Run `/bitwarden-init:enhance-user` — confirm it only offers modules that are NOT already present in the file
- [ ] From a directory under a folder containing 2+ cloned repos, run `/bitwarden-init:init-workspace` — confirm the auto-detection proposes the right root and that the diff preview appears before any write
- [ ] Confirm `/bitwarden-init:init` and `/bitwarden-init:enhance` (repo scope) still behave exactly as before


plugins/bitwarden-init/commands/init-workspace/init-workspace.md
```diff
@@ -0,0 +1,91 @@
...
+
+This represents one engineer's opinion. Modules are opt-in; the user picks which to include and is free to edit afterward.
+
+## Steps
+
+1. **Detect workspace root.** Starting from `cwd`, walk up the directory tree. For each ancestor, count how many of its immediate subdirectories contain a `.git` entry (file or directory). If any ancestor has **2 or more** such subdirectories, propose that ancestor as the workspace root.
+
+   Use a Bash one-liner like:
+
+   ```bash
```
theMickster, line 16: ♻️/🎨 A local multi-agent code review brought up the following finding. I think it's worth considering

<details>

<summary>Workspace-root detection walks all the way to `/` with no `$HOME` guard</summary>

`plugins/bitwarden-init/commands/init-workspace/init-workspace.md:18-26`

  The detection loop has no upper bound at `$HOME`:

  ```bash
  dir="$PWD"
  while [ "$dir" != "/" ]; do
    count=$(find "$dir" -mindepth 2 -maxdepth 2 -name .git 2>/dev/null | wc -l | tr -d ' ')
    if [ "$count" -ge 2 ]; then echo "$dir"; break; fi
    dir=$(dirname "$dir")
  done
  ```

  Engineers who keep any two git repos directly under their home directory (dotfiles, config repos, personal projects) will always see `~` proposed as the workspace root. That resolves in step 2 to `~/.claude/CLAUDE.md`, which collides with the user-scope file that `init-user` manages. The explicit confirmation prompt (step 2) prevents an accidental write, but the misleading proposal fires every time.

  **Suggested fix:** Add `if [ "$dir" = "$HOME" ]; then break; fi` at the top of the loop body. Document this constraint in the command Notes.
  </details>
</details>

plugins/bitwarden-init/commands/enhance-workspace/enhance-workspace.md
```diff
@@ -0,0 +1,44 @@
...
+
+Refresh an existing workspace-level CLAUDE.md by detecting which Bitwarden domain modules are already present and offering to add the missing ones. Same safety guardrails as `/bitwarden-init:init-workspace` — backup before overwrite, diff preview, explicit Apply confirmation.
+
+## Prerequisites
+
+A workspace-level CLAUDE.md must already exist. If it doesn't, stop and point the user at `/bitwarden-init:init-workspace` instead.
+
+## Steps
+
+1. **Resolve workspace root** the same way `/bitwarden-init:init-workspace` does:
```
theMickster: ♻️ / 🎨 We have near workspace root detection algorithm here. Should we consider something like:

```suggestion
Follow the workspace-root detection algorithm from `/bitwarden-init:init-workspace` step 1 exactly, including the confirmation prompt.
```

**-OR-**

Do we limit the complexity and instead leverage the AskUserQuestion tool for the path of the workspace root? Simplifies the flow and long-term maintenance burden.

plugins/bitwarden-init/commands/init-user/init-user.md
```diff
@@ -0,0 +1,80 @@
...
+
+   Read the current plugin version from `${CLAUDE_PLUGIN_ROOT}/.claude-plugin/plugin.json` to substitute `{VERSION}`. Use today's date for `{YYYY-MM-DD}`.
+
+4. **Diff + confirm.**
+   - If the target file exists, write the rendered content to a temp file (e.g. `/tmp/claude-md-preview-$$.md`) and run `diff -u "$HOME/.claude/CLAUDE.md" /tmp/claude-md-preview-$$.md` via `Bash`. Display the diff to the user.
+   - If the target does not yet exist, display the full rendered content.
+
+   Then use `AskUserQuestion` with options **Apply**, **Show diff again**, **Cancel**. On **Show diff again**, redisplay and ask again. On **Cancel**, stop without writing.
+
+5. **Backup, then write.**
```
theMickster, line 64: ♻️ / 🎨 A multi-agent code review found the following finding. I worked with Claude to double-check it's work regarding it's claims about how the Write and Edit tools work by referencing actual Anthropic documentation. 

<details>
<summary>
`init-user.md` step 5 uses Write-primary / Edit-fallback, opposite of `enhance-user.md`
</summary>
<br/>
`plugins/bitwarden-init/commands/init-user/init-user.md:64-67`<br/>
`plugins/bitwarden-init/commands/enhance-user/enhance-user.md:31-34`<br/><br/>
**Caught by:** Code quality agent<br/><br/>

`init-user.md` step 5 (line 66): *"Use `Write` to create `~/.claude/CLAUDE.md` with the rendered content. (If `Write` rejects modifying a file that exists, fall back to `Edit` with the full-file replacement.)"*

`enhance-user.md` step 6 (line 33): _"Use `Edit` to update `~/.claude/CLAUDE.md` with the full rendered content. (`Write` works as a fallback if `Edit`'s old-string match fails for any reason — the file always exists at this point.)"_

The fallback condition is dead logic. Per the [Write tool docs](https://code.claude.com/docs/en/tools-reference#write-tool-behavior), `Write` only rejects an existing file if it hasn't been read — but step 1 already reads the file when it exists, so `Write` cannot be rejected here. The correct ordering is `Edit` primary (per the [Edit tool docs](https://code.claude.com/docs/en/tools-reference#edit-tool-behavior), a string-match miss is a real failure mode that warrants a fallback) with `Write` as the unconditional fallback — exactly what `enhance-user.md:33` does. However, `init-user.md` also has a new-file path (step 1 found nothing) where `Edit` is not viable: Edit requires read-before-edit, and a non-existent file was never read.

**Suggested fix:** Split step 5 by path. When the file exists (read in step 1): `Edit` primary, `Write` as fallback on string-match failure. When creating a new file: `Write` only. `enhance-user.md` uses Edit-primary unconditionally because its prerequisite guarantees the file exists; `init-user.md` cannot make that assumption.

</details>

plugins/bitwarden-init/templates/user/no-comments-default.md
```diff
@@ -0,0 +1,9 @@
+## Default to [YOUR-PREFERENCE: No Comments | Documented Comments]
```
theMickster: 🎨 / 💡 Should we consider the following minor change? I wonder if Claude code will correctly resolve the placeholders and, instead, prompt the user to add the block a second time if not found. Perhaps something like the following will be more reliably matched. If so then we would need to make minor alterations to the `enhance-user` and `init-user` commands.

```suggestion
## Comment Policy 

### Default to [YOUR-PREFERENCE: No Comments | Documented Comments]
```

plugins/bitwarden-init/commands/init-workspace/init-workspace.md
```diff
@@ -0,0 +1,91 @@
+---
+description: Set up a workspace-level CLAUDE.md (Bitwarden domain context) for the folder containing all locally-cloned Bitwarden repos
+allowed-tools: Read, Write, Edit, Bash(find:*), Bash(wc:*), Bash(tr:*), Bash(dirname:*), Bash(mkdir:*), Bash(diff:*), Bash(cp:*), Bash(date:*), Bash(rm:*), Glob
```
theMickster, line 3: 🎨 These tools feel overly broad to me. Shall we consider limiting them in scope to the temporary files/folders or leaving them off altogether. 

I would support prompting the user for acceptance of tool usage because the process isn't a _set it and forget it_ workflow; it's an interactive back-n-forth with Claude IMO.

plugins/bitwarden-init/commands/enhance-workspace/enhance-workspace.md
```diff
@@ -0,0 +1,44 @@
+---
+description: Refresh an existing workspace-level CLAUDE.md by adding missing Bitwarden domain modules
+allowed-tools: Read, Write, Edit, Bash(find:*), Bash(wc:*), Bash(tr:*), Bash(dirname:*), Bash(diff:*), Bash(cp:*), Bash(date:*), Bash(rm:*), Glob
```
theMickster: 🎨 These tools feel overly broad to me. Shall we consider limiting them in scope to the temporary files/folders or leaving them off altogether. 

I would support prompting the user for acceptance of tool usage because the process isn't a _set it and forget it_ workflow; it's an interactive back-n-forth with Claude IMO.

plugins/bitwarden-init/commands/enhance-user/enhance-user.md
```diff
@@ -0,0 +1,41 @@
+---
+description: Refresh an existing user-level ~/.claude/CLAUDE.md by adding missing modules from the Bitwarden init template library
+allowed-tools: Read, Write, Edit, Bash(diff:*), Bash(cp:*), Bash(date:*), Bash(rm:*), Glob
```
theMickster, line 3: 🎨 These tools feel overly broad to me. Shall we consider limiting them in scope to the temporary files/folders or leaving them off altogether. 

I would support prompting the user for acceptance of tool usage because the process isn't a _set it and forget it_ workflow; it's an interactive back-n-forth with Claude IMO.

plugins/bitwarden-init/commands/init-user/init-user.md
```diff
@@ -0,0 +1,80 @@
+---
+description: Set up a user-level ~/.claude/CLAUDE.md with opt-in Bitwarden behavioral modules through a guided Q&A
+allowed-tools: Read, Write, Edit, Bash(diff:*), Bash(cp:*), Bash(date:*), Bash(rm:*), Glob
```
theMickster, line 3: 🎨 These tools feel overly broad to me. Shall we consider limiting them in scope to the temporary files/folders or leaving them off altogether. 

I would support prompting the user for acceptance of tool usage because the process isn't a _set it and forget it_ workflow; it's an interactive back-n-forth with Claude IMO.

theMickster (commented):
Couple items for discussion.

plugins/bitwarden-init/commands/init-workspace/init-workspace.md
withinfocus: 🎨 Going to push back on this one — happy to revisit if I'm missing an attack vector.

The current `allowed-tools` is already narrowed at the subcommand level (`Bash(diff:*)`, `Bash(cp:*)`, `Bash(date:*)`, `Bash(rm:*)`, plus `find`/`wc`/`tr`/`dirname`/`mkdir` for the workspace flows). Two reasons not to tighten further:

**Path-pattern restrictions** would have to encode `/tmp/claude-md-preview-$$.md`, where `$$` is the process pid — so the path varies per invocation. A wildcard like `Bash(rm:/tmp/claude-md-preview-*)` catches our temp file but would also match any other tool's `claude-md-preview-*` debris. The sticky validator on this same PR flagged that tightening as over-engineering, and I agree.

**Removing the tool entries entirely** would force a permission prompt for every `diff -u`, `cp`, `date`, `rm` invocation — several per run, all on innocuous calls. The Q&A already has an explicit `Apply`/`Cancel` confirmation via `AskUserQuestion` before any mutation, with a unified diff preview right above it. That's the gate that semantically matters: the user sees exactly what content goes where. Adding per-Bash-call prompts on top would drown that signal in noise — the user sees "may we run `date`?" four times before the one prompt that actually matters, and after a few runs the `date` prompts train them to click through.

I think the Apply gate is the right level of friction, but if there's a concrete scenario where a Bash subcommand we whitelisted could be abused inside this flow, I'd love to hear it — happy to be wrong.

plugins/bitwarden-init/commands/enhance-workspace/enhance-workspace.md
withinfocus: Same answer as on the init-workspace.md thread — see https://github.com/bitwarden/ai-plugins/pull/119#discussion_r3333015304 for the rationale. Happy to revisit if there's a concrete attack vector the `Apply` gate doesn't catch.

plugins/bitwarden-init/commands/enhance-user/enhance-user.md
withinfocus: Same answer as on the init-workspace.md thread — see https://github.com/bitwarden/ai-plugins/pull/119#discussion_r3333015304 for the rationale. Happy to revisit if there's a concrete attack vector the `Apply` gate doesn't catch.

plugins/bitwarden-init/commands/init-user/init-user.md
withinfocus: Same answer as on the init-workspace.md thread — see https://github.com/bitwarden/ai-plugins/pull/119#discussion_r3333015304 for the rationale. Happy to revisit if there's a concrete attack vector the `Apply` gate doesn't catch.

plugins/bitwarden-init/commands/init-workspace/init-workspace.md
theMickster: Makes sense. Appreciate the discussion/thought pattern

theMickster (approved):

PR #121, opened 2026-05-14
fix(delivery-tools): harden creating-pull-request workflow and trigger reliability

SaintPatrck:
## 🎟️ Tracking

PM-37558

## 📔 Objective

Harden `bitwarden-delivery-tools:creating-pull-request` against three failure modes:

1. **PR template skipped / label dropped at submission.** Workflow is rewritten as six ordered steps with a mandatory Step 5 submission preview (target repo, branch, title, type prefix → t: label, ai-review label, body) shown before any `git push` or `gh pr create`. Step 5 confirmation now uses `AskUserQuestion` with structured options (Submit as shown / Edit title or body / Change ai-review label / Cancel), mirroring Step 4 and removing the parse ambiguity of a free-text question.

2. **Description under-triggered on natural language.** The previous description was narrow ("create PR", "open PR", "gh pr create") and missed casual phrasings. Rewrote it to cover "package this up into a PR", "ship a draft", "send it up for review", "get this in front of reviewers", "wrap this branch", "throw together a PR", etc., and asserted precedence over `labeling-changes` when the user is in a PR-creation context. Explicit exclusions for conceptual queries ("explain how PRs work") and existing-PR management (status, merging, addressing comments) keep precision intact.

3. **Frontmatter verbosity.** Split the previously-single ~1,485-char description into `description` (what the skill does) and `when_to_use` (when to invoke it) per the [Claude Code skills frontmatter reference](https://code.claude.com/docs/en/skills#frontmatter-reference) and the [agentskills.io optimizing-descriptions guide](https://agentskills.io/skill-creation/optimizing-descriptions#writing-effective-descriptions). Combined listing chars dropped to ~1,137 (cap 1,536).

### Validation

Added `plugins/bitwarden-delivery-tools/skills/creating-pull-request/evals/`:

- `trigger-eval.json` — 20-query test set (10 should-trigger natural-language phrasings + 10 should-not-trigger near-misses against sibling delivery skills).
- `run_real_eval.py` — custom runner that works around an upstream skill-creator harness false-negative when the skill under test is already plugin-registered.
- `baseline.json` — known-good run for diff-based regression checks on future description changes.

Measured 100% trigger reliability by threshold (3 runs/query, claude-opus-4-7) across multiple consecutive iterations during development. Replace `baseline.json` whenever a future change re-passes; the rest is durable.


.claude-plugin/marketplace.json
```diff
@@ -73,7 +73,7 @@
     {
       "name": "bitwarden-delivery-tools",
       "source": "./plugins/bitwarden-delivery-tools",
-      "version": "1.1.0",
+      "version": "1.2.0",
```
withinfocus: 💭 I don't really know what the versioning rules are for the whole marketplace.

plugins/bitwarden-delivery-tools/skills/creating-pull-request/SKILL.md
```diff
@@ -1,70 +1,139 @@
...
+<!-- Required for UI changes; delete if not applicable. -->
 ```
 
-Delete the Screenshots section entirely if there are no UI changes.
+### Step 4 — Ask about the AI review label
+
+Use the `AskUserQuestion` tool to ask:
+
+- **Question**: "Would you like to add an AI review label to this PR?"
+- **Options**: `ai-review-vnext`, `ai-review`, `No label`
```
aj-bw: @SaintPatrck I was running this locally and got the options as expected! do you think it would be good to swap `ai-review` to be first so the vnext is more opt in?

plugins/bitwarden-delivery-tools/skills/creating-pull-request/evals/run_real_eval.py
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env python3
```
withinfocus, line 1: ⚠️ I think this is our first Python file, so please amend the `.editorconfig` file to ensure formatting.

plugins/bitwarden-delivery-tools/skills/creating-pull-request/SKILL.md
```diff
@@ -1,70 +1,139 @@
...
-- `[PM-12345] refactor: Simplify authentication flow`
+**Show the proposed title to the user before continuing.** This is the first chance for them to catch typos, a missing prefix, or the wrong ticket key.
 
----
+### Step 3 — Read the repo's PR template
+
+Always read `.github/PULL_REQUEST_TEMPLATE.md` from the target repo before drafting the body. Even when you have a body draft in mind, the template's sections are what other reviewers expect to scan. Skipping this is a common failure mode — PRs ship with improvised bodies that miss sections reviewers depend on.
 
-## PR Body
+If the template exists:
```
withinfocus, line 49: ℹ️ This really always should, but it makes me wonder how often this fails to find it and then doesn't use the in-repo one, silently.

withinfocus (changes requested):
One important maintenance item.

Is it worth it at this point to pull the eval scripts and execution to a common location, so that it can be used by other plugins as well? Can we make it generic for repo-wide usage?

plugins/bitwarden-delivery-tools/skills/creating-pull-request/SKILL.md
SaintPatrck: Yea, that sounds good. 👍

plugins/bitwarden-delivery-tools/skills/creating-pull-request/SKILL.md
SaintPatrck: I originally added it as an extra safe-guard to make sure we see something at least slightly recognizable. Happy to reword so that failures are surfaced, if they occur, so we can investigate root causes. LMK.

.claude-plugin/marketplace.json
SaintPatrck: Good docs gap callout. I can open a followup to add details.

To answer directly, the version is essentially a cache key, so we have to bump it on every change for consumers to get them.

We have two options. We either update `version` every time a change is made, or we can remove `version` entirely. If we remove `version` the commit SHA becomes the cache key and Claude will pull down changes on every new commit to the marketplace/plugin(s). 

Version feels like the right approach because it's pinned. That being said, we don't version pin other internal tools, so I'm open to suggestion if there's strong feelings one way or the other.

All of this holds true for the marketplace version and the plugin versions.

plugins/bitwarden-delivery-tools/skills/creating-pull-request/evals/run_real_eval.py
SaintPatrck: It is. I found some rules in our splunk repo. Are there any python specific rules beyond `indent_size = 4` that we want to specify?

SaintPatrck:
>Is it worth it at this point to pull the eval scripts and execution to a common location, so that it can be used by other plugins as well? Can we make it generic for repo-wide usage?

I think we should consider that a separate effort to avoid increasing scope of this PR. I can follow-up with another PR that generalizes it and adds an accompanying `evals.json` in another plugin to verify generalizing it makes sense.

.claude-plugin/marketplace.json
withinfocus: Consumers to get what though? New plugins, or updates to existing ones?

If we don't use the pin on the consumption side then I don't see a need to maintain this though -- we want our users to pick up any updates pretty seamlessly.

plugins/bitwarden-delivery-tools/skills/creating-pull-request/evals/run_real_eval.py
withinfocus: I don't think so.

withinfocus (dismissed):
Changelog is a bit lengthy but this works and addressed what I thought was essential. Couple things for future PRs.

SaintPatrck:
> Changelog is a bit lengthy but this works and addressed what I thought was essential. Couple things for future PRs.

Agreed. Updated to be more concise.

withinfocus (approved):

PR #138, opened 2026-06-07
Reframe tech-lead/software-engineer as engineering-shaping/engineering-building

trmartin4:
## Objective

Two role-named agents (`bitwarden-tech-lead` and `bitwarden-software-engineer`) carved the engineer's reponsibilitites along a role axis ("tech lead role" vs. "engineer role"). In practice, the same engineer slides between **shaping** (discovery, scoping, architectural reasoning, cross-team coordination, tech-breakdown drafting) and **building** (implementing stories, reviewing PRs, preparing commits, shipping). The role-based split forced users to pick a hat that didn't match what they were actually doing.

This PR reframes the two plugins along activity mode instead of role:

- `bitwarden-tech-lead` → **`bitwarden-engineering-shaping`** (the work of forming workable shape from vague intent)
- `bitwarden-software-engineer` → **`bitwarden-engineering-building`** (the work of constructing the result)

Both are for any engineer; the split is by what you're doing, not what title you hold. The skills (`architecting-solutions`, `contributing-to-technical-strategy`, the delivery-tools workflow skills, etc.) stay agent-neutral and remain invocable from either agent.


theMickster:
I am struggling to see why the name change is a now priority compared to building the Claude tooling leveraged by these agents. Pure cost-benefit for me.  I don't really mind what the agents are called, but more so that they have the skills, rules, etc. to aiding engineering.   I think that's where we need to be investing our efforts.

I could also be missing a bigger picture missing in the Claude generated description. How does the change help us achieve the bigger VFO for AI + Engineering excellence?  What planned Claude tooling enhancements will directly benefit from this change that I don't see?

PR #141, opened 2026-06-09
Breakdown initialization and spec definition skills

trmartin4:
## 📔 Objective

The first in a series of changes to decompose and enhance the existing `writing-tech-breakdowns` skill, for use with our new `tech-breakdowns` repo.

The plan for skill decomposition is roughly defined below, and these are the first 2 skills in that plan:
- [X] Creating a new breakdown (this PR)
- [X] Defining the Specification (the WHAT and WHY) (this PR)
- [ ] Defining the Plan (the HOW) https://github.com/bitwarden/ai-plugins/pull/143
- [ ] Defining the Tasks https://github.com/bitwarden/ai-plugins/pull/144
- [ ] Completing the breakdown (archiving to avoid ambiguous future references)

Depends upon https://github.com/bitwarden/tech-breakdowns/pull/10 for naming of `breakdown.md` template.

For current ideation on all the remaining skills, you can reference #134, which will not be merged in its current state

PR #134, opened 2026-06-04
Skill updates for new breakdown repo

trmartin4:
## 📔 Objective

The breakdown-related skills are re-anchored to the new [`bitwarden/tech-breakdowns`](https://github.com/bitwarden/tech-breakdowns) repo (single markdown file per breakdown under the owning team's folder), and the monolithic `writing-tech-breakdowns` skill is replaced with phase-scoped skills that match how engineers actually do the work: orient, specify, plan, decompose, sync to Jira. Each phase is a small, composable skill.  The rough framing of the Github `spec-kit` model was used to identify axes of separation.

---

### Removed: `writing-tech-breakdowns` skill

A single skill covered drafting, status transitions, cross-team engagement, signoff table, Jira story creation, and gates. It was too large to reliably route mid-stream prompts (lost-in-the-middle), and it tangled policy with operational mechanics.

Replaced by the five phase-scoped skills below.

### Removed: `coordinating-cross-team-breakdown` skill

Cross-team engagement (identifying affected teams, characterizing impacts by depth and owning-team churn, building the signoff table) now happens inside `developing-the-plan` activity 5. There is no separate signoff-chasing skill — reviewers fill the Signoff column during cross-team review between `Proposed` and `Accepted`, and the table fills itself as they respond.

---

## New phase-scoped tech-breakdown skills

The lifecycle is broken into five skills, aligned with [Spec-Kit](https://github.com/github/spec-kit)'s `/specify` → `/plan` → `/tasks` decomposition:

### `starting-a-tech-breakdown`

Responsible for file setup and mechanics.  Open-ended context gathering (prompts the user for Jira key, where existing context lives, and the named owner), copies the template, and fills the Status block.

### `understanding-the-work` — orient and resolve open design questions

Walk the engineer through understanding the change well enough to start specifying it. Three phases: Resume (skip if fresh), Orient (Decided / Open / Gaps summary), Resolve (one question at a time with 2–3 concrete options). The activity here is the thinking; we aren't yet focused on the "how", but on the "what" and the "why". Uses the Clarifications Log as **dual-use working state and reviewer artifact** (Anthropic's structured-note-taking pattern): capture liberally during resolution; the planning skill curates before reviewer-ready. `HARD-GATE` blocks advancing while Open clarifications remain.

### `developing-the-spec` — Spec-Kit `/specify` analog

[...]

PR #143, opened 2026-06-10
Breakdown plan skill

trmartin4:
## 📔 Objective

Adds the next in a sequence of changes to decompose and enhance the existing writing-tech-breakdowns skill, for use with our new tech-breakdowns repo.

The plan for skill decomposition is roughly defined below, and this is the 3rd skill in that plan:

The plan for skill decomposition is roughly defined below, and these are the first 2 skills in that plan:
- [X] Creating a new breakdown https://github.com/bitwarden/ai-plugins/pull/141
- [X] Defining the Specification (the WHAT and WHY) https://github.com/bitwarden/ai-plugins/pull/141
- [X] Defining the Plan (the HOW) (**this PR**)
- [ ] Defining the Tasks https://github.com/bitwarden/ai-plugins/pull/144
- [ ] Completing the breakdown (archiving to avoid ambiguous future references)

For current ideation on all the remaining skills, you can reference #134, which will not be merged in its current state


plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+- Specification is empty or partial — prompt the user to define the Specification before continuing. The Plan needs the Spec as its anchor; without one, the Plan has no constraint to design against.
+- Open design questions remain in the Clarifications Log. Instruct the user to resolve them first.
+
+</HARD-GATE>
+
+## Key Principles
+
+- **Spec anchors the Plan.** No Plan content while the Spec is empty or partial.
+- **Verify before claiming.** Read the file or grep before saying "the code does X"; never assume based on a description.
+- **Link, don't duplicate.** If a decision is documented in a PRD, Jira issue, or Slack thread, reference it.
```
theMickster: Couple things:
💡 I recommend expanding the acronym PRD here because it's the first time it's referenced (I assume this is a Product Requirements Document?).  Although Claude will likely get this right, I always recommend this as a best practice in technical writing. 
❓In PR #141, we have called out something similar _The PRD or Architecture Plan_ as part of `developing-breakdown-spec`. Does it matter to be consistent here in calling for Claude to help the human link to an Architecture Plan as well?
❓Will Claude have access to Slack & PRD documents? How will it know what to link? Or is this principle here to ensure Claude guide's the human to link and not duplicate?

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+If both gates pass, triage which activities (below) are complete and which remain. Continue with the next unfinished one.
+
+### Developing the Plan
+
+Work through these activities. Order is sequential — each depends on the previous — and the self-review at the end is explicitly the last step.
+
+#### 1. Develop the technical architecture to meet the Specification
+
+- Invoke `Skill(architecting-solutions)` first to apply the architectural lens.
+- Route any cryptographic work through `Skill(bitwarden-security-context)`.
```
theMickster: 🎨 Let's consider refactoring using the verb Invoke just like we did above. 

```suggestion
- Invoke `Skill(bitwarden-security-context)` for planning any cryptographic work.
```

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+
+#### 2. Map per-layer impact
+
+Walk every per-layer area the change touches — DB, server, clients, SDK, mobile, infrastructure, anything else. Use the checklist in each section of the breakdown to ensure that all potential impacts on each layer are addressed.
```
theMickster: 💡The [tech-breakdown.md](https://github.com/bitwarden/tech-breakdowns/blob/main/templates/tech-breakdown.md) template lists these layers in a different detail that may confuse Claude. We may want to consider refactoring this sentence to anchor/ground Claude to the exact places in the template you want to to start and then end.  

Something like the following might be better to keep Claude precisely on task.
> starting with `## Data model changes` and working all the way through `## Client / UI behavior changes`

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+**B. Characterize the impact across two inputs.** Don't skip either; if unknown, name it as unknown so the assessment is conditional.
+
+1. **Domain-overlap depth** — _Surface_ (mechanical, well-documented patterns, no domain reasoning), _Mid_ (must follow established contracts, naming, error-handling conventions), _Deep_ (touches the owning team's core invariants, mental model, or design rationale).
+2. **Owning-team domain churn** — is the owning team actively reshaping the area? **Scan explicitly; don't guess.** Three surfaces:
+   - **In-flight breakdowns in the owning team's folder of `bitwarden/tech-breakdowns`**, excluding `**/complete/**`:
+     ```
+     grep -rli "<repo-name>" <owning-team>/ --include="*.md" --exclude-dir=complete
+     grep -rli "<file-or-module-name>" <owning-team>/ --include="*.md" --exclude-dir=complete
+     ```
+     Read candidate breakdowns' Tasks and Plan sections to confirm overlap rather than relying on grep matches alone.
```
theMickster, line 108: ❓Did something get sideways in the markdown? Strange that GitHub is showing everything down from line 101 in blue like it's a code block, eh? 🤔

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+---
+
+# Developing the Plan
+
+## Overview
+
+Assist a Bitwarden engineer in developing the HOW a change will be built, anchored to the already-defined Specification section of the breakdown document. The skill iterates on a technical architecture with the user, walks the change against every part of our technical stack to surface impact, scans for in-flight work that could collide, identifies and characterizes every cross-team impact, and runs a final self-review pass against the breakdown template.
+
+<HARD-GATE>
+Orientation within a breakdown is required. Ask the user which breakdown to work against. They can give a path, a Jira key, or a team/slug — use `Glob` under `bitwarden/tech-breakdowns/` to resolve to a real `breakdown.md`. If the user already named it earlier in the conversation, confirm the resolved path with `AskUserQuestion` before proceeding.
```
theMickster: ⚠️ I don't think that the `bitwarden/tech-breakdowns` will work here because we're assuming that everyone has their workstation configured exactly like yours with a bitwarden folder. 

Also worth mentioning that the gate/instructions aren't particularly clear where the end user should start their Claude session to use the skill. It's clear that we're working on tech-breakdowns, but we also are asking Claude (and the human) to search & link to artifacts. Would the human be better off being told to start the process in the root folder where he/she has cloned all the Bitwarden repos so that Claude can help traverse those repos to create the Plan?

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+
+_Captured in **Cross-team engagement** (Consuming other teams' APIs, Changes required in other teams' code, Cross-team sequencing & ordering, plus the signoff table and Coordination notes)._
+
+For an end-to-end illustration of the (A) → (B) → (C) walk for one realistic impact, see `references/worked-example.md`.
+
+#### 5. Self-review the breakdown
+
+Final pass before the breakdown is reviewer-ready. Run it yourself against the saved file; no subagent. If you find issues, fix them inline and move on.
+
+1. **Template-section coverage** — open the breakdown template (`bitwarden/tech-breakdowns/templates/tech-breakdown.md`) and confirm every top-level and subsection from the template appears in the breakdown, with either real content or an explicit `N/A — <reason>`. Empty section bodies are a finding; resolve before continuing.
```
theMickster: ⚠️ / ♻️ Couple items to consider. Both were called out during a local [performing-multi-agent-code-review](https://github.com/bitwarden/ai-plugins/tree/main/plugins/bitwarden-code-review/skills/performing-multi-agent-code-review).

1. The pathing defined here is likely not going to work outside of your workstation because of the `bitwarden` folder.  Also, this pairs with my comment around line 14 because we don't instruct at what level a human should start a Claude session at when developing the plan.
2. We have a disconnect here where we call out `tech-breakdown.md` instead of `breakdown.md`.  We refer to `breakdown.md` in sibling skills and above around line 14. Please review all instances of the file name and align.

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+Walk every per-layer area the change touches — DB, server, clients, SDK, mobile, infrastructure, anything else. Use the checklist in each section of the breakdown to ensure that all potential impacts on each layer are addressed.
+
+Be specific, and address the checklist items in each of the sections. Plan is where the concrete file and module list emerges, and downstream activities need an accurate list to act on. _Captured in **Plan**._
+
+#### 3. Scan for in-flight work
+
+Now that the Plan has produced a concrete file and module list, scan three sources for work that could collide:
+
+- **Other teams' breakdowns** in `bitwarden/tech-breakdowns`, excluding `**/complete/**`. Grep for the affected file paths and module names across the tree.
+- **Open PRs in the affected repos**: `gh pr list -R bitwarden/<repo> --state open --json number,title,headRefName,files`. Look for PRs touching the same files.
```
theMickster, line 82: ♻️ A local [performing-multi-agent-code-review](https://github.com/bitwarden/ai-plugins/tree/main/plugins/bitwarden-code-review/skills/performing-multi-agent-code-review) called this out and I think we need to consider it as a security related refactoring. 

<details>
<summary>In-flight-scan commands interpolate skill-classified-untrusted names into the shell without a safe-handling directive</summary>

<br/>`plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md:77-99`
<br/>**Caught by:** Security & logic agent — *downgraded from Important to Refactor on verification*<br/>

Activity 3 and Activity 4 show example commands that substitute placeholders derived from breakdown content the skill itself labels untrusted (Key Principles bullet, line 28): `gh pr list -R bitwarden/<repo> ...` and `git log ... -- <path>` (lines 77-78), and `grep -rli "<file-or-module-name>" <owning-team>/ ...` (lines 98-99). The untrusted-data principle only tells the agent not to *follow instructions* found in fetched content; it gives no directive to safely handle those values when interpolated into a shell command, and the grep example's double quotes do not neutralize `$(...)`, backticks, or `";cmd"`.

**Honest assessment (why this is Refactor, not Important):** exploitation requires a speculative input — an attacker planting shell metacharacters into a *file path or module name* inside the internal `bitwarden/tech-breakdowns` repo, which an agent then interpolates unquoted. The values are real filesystem paths/module names, the repo is internal, and the skill drives an interactive agent rather than an automated script generator. The skill's own "Do Not Flag" rule excludes speculative issues without evidence the inputs occur in practice, so this does not clear the Important ("likely hit in practice") bar. The defensible kernel is a consistency gap: the skill recognizes the content as untrusted but does not extend that to shell handling. No vault data, keys, or auth tokens (VD/EK/AT) are implicated; any risk is bounded to the engineer's local environment.

Suggested fix (defensive hardening): bind untrusted-derived values as fixed-string, non-shell-interpreted arguments — e.g. `grep -F -- "$NAME"` — and add a one-line directive: "never splice breakdown-derived names into a shell-evaluated command string; bind them as literal arguments."
</details>

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+2. Present 2 or 3 concrete options with tradeoffs. If you can't articulate at least two, surface that as a finding.
+3. Verify against actual code or docs when the question turns on what exists.
+4. Wait for the user's decision.
+5. Record it in the Clarifications Log as `Resolved`, with owner and date.
+
+## Workflow
+
+Ask the user up front: starting a new Plan, or continuing one? If continuing, work through **Resuming a Plan** first, then **Developing the Plan**. If starting new, go straight to **Developing the Plan**.
+
+Create a task for each section as you start it (`TaskCreate`), mark it in progress, and complete it before moving on. If resuming, use `AskUserQuestion` to confirm which activity to pick up at and re-fetch external sources (Jira, PRD, PoC) before continuing. See `references/process-flow.dot` for the full decision graph.
```
theMickster: ♻️ We instruct Claude to resume the process by fetching outside sources, but do not provide tool access in `allowed-tools`. Please consider either of the following:

1. **Fix (additive):** Add our Atlassian MCP tools to `allowed-tools`, matching `developing-breakdown-spec/SKILL.md`.
2. **Fix (clarifying):** Change the instruction to "re-read the breakdown document to reload context" and remove the "(Jira, PRD, PoC)" enumeration if the intent is to rely on what is already captured in the Spec section.

plugins/bitwarden-tech-lead/CHANGELOG.md
```diff
@@ -5,6 +5,24 @@ All notable changes to the `bitwarden-tech-lead` plugin will be documented in th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.3.1] - 2026-06-10
```
theMickster, line 8: ⚠️ Please condense this into a simple changed block; these are overkill. 
These entires Claude made are as long as the changes themselves 🤣

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/references/worked-example.md
theMickster, line 1: ❓I'm a little confused here by the purpose of the worked-example. It doesn't look like the output of the `developing-breakdown-plan` which (from what I understand) is a completed `# Plan` section of a `tech-breakdown.md` file. Correct? 

I think this is a classic overstatement by Claude's `/skill-creator` plugin telling us humans that these are a required best practice. I disagree. These are excess fluff to compensate for a poorly written skill or poor context. That's is most definitely not what we have in this PR; so I recommend we remove it until we have concrete evidence that Claude cannot use the information found in the template checklist & `SKILL.md` to create the Plan. 

Let's be real, the checklists in the template itself is arguably the strongest anchor for Claude 🤔 🤷🏼‍♂️

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/references/process-flow.dot
```diff
@@ -0,0 +1,41 @@
...
+  LayerImpact    [shape=box,     label="Activity 2:\nMap per-layer impact"];
+  InFlight       [shape=box,     label="Activity 3:\nScan for in-flight work"];
+  CrossTeam      [shape=box,     label="Activity 4:\nIdentify cross-team impacts"];
+  SelfReview     [shape=box,     label="Activity 5:\nSelf-review\n(template coverage,\nspec coverage,\nplaceholders, consistency)"];
+  ReviewerReady  [shape=diamond, label="Reviewer-ready?"];
+  Done           [shape=ellipse, label="Done →\nteam-internal review,\nthen move to Proposed"];
+
+  Start         -> EntryMode;
+  EntryMode     -> GatesPass     [label="fresh"];
+  EntryMode     -> Triage        [label="resume\n(re-read file,\nre-fetch sources)"];
```
theMickster: ⚠️ / ♻️ A two local multi-agent reviews plus a third independent session running argumentative agents (I enjoy have Claude argue against itself 🤭) came up with the following finding for review. 
Seems reasonable given all of the potential points to resume the Plan section at, but, I've not worked with these types of documents before so deferring to you, Todd. Thoughts?

<details>

<summary>Resume path: Triage runs before the gate check and has no mid-flow dispatch edges</summary>

<br/>plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/references/process-flow.dot:30-31
<br/>Caught by: Adversarial panel (all three agents; Owl identified both symptoms as one root cause)<br/>

Two interlocked defects on the resume entry path.

Symptom 1 — inverted gate order (line 30): The resume path is EntryMode -> Triage -> GatesPass. SKILL.md §Resuming is explicit: verify both gates first ("Spec filled? Clarifications resolved?"), and only if they pass does triage occur. The diagram inverts this. An agent following the graph surfaces triage output — a summary of which activities are complete and which remain — before knowing whether continuation is even permitted. If the gate then fails, that triage output is noise produced in a state where the prose says to stop and return to Spec.

Symptom 2 — Triage discards its own result (line 31): Triage's only outbound edge is Triage -> GatesPass. After GatesPass -> Architecture, the chain runs all five activities in fixed order. The file header explicitly promises "the user picks up from whichever activity remains unfinished," and SKILL.md §Resuming says "Continue with the next unfinished one." But the graph has zero edges from Triage to LayerImpact, InFlight, CrossTeam, or SelfReview. A Triage node that computes which activities are complete and then routes unconditionally to Activity 1 produces a computation it immediately discards. Every resume restarts at Activity 1 and re-runs all five activities — including re-invoking Skill(architecting-solutions) on already-completed architecture work.

Both symptoms share a fix: gate must precede triage, and triage must fan out to each activity.

Fix:
EntryMode     -> GatesPass    [label="resume\n(re-read file,\nre-fetch sources)"];
GatesPass     -> Triage       [label="yes (resume)"];
Triage        -> Architecture  [label="resume at 1 (if unfinished)"];
Triage        -> LayerImpact   [label="resume at 2 (if unfinished)"];
Triage        -> InFlight      [label="resume at 3 (if unfinished)"];
Triage        -> CrossTeam     [label="resume at 4 (if unfinished)"];
Triage        -> SelfReview    [label="resume at 5 (if unfinished)"];
Update the header comment to reflect gate-first order.

</details>

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/references/process-flow.dot
```diff
@@ -0,0 +1,41 @@
...
+  Triage        -> GatesPass;
+  GatesPass     -> StopGate      [label="no"];
+  GatesPass     -> Architecture  [label="yes"];
+  Architecture  -> LayerImpact;
+  LayerImpact   -> InFlight;
+  InFlight      -> CrossTeam;
+  CrossTeam     -> SelfReview;
+  SelfReview    -> ReviewerReady;
+  ReviewerReady -> Architecture  [label="no, next\nunfinished activity"];
```
theMickster: ⚠️ / ♻️ A two local multi-agent reviews plus a third independent session running argumentative agents (I enjoy have Claude argue against itself 🤭) came up with the following finding for review. 
This one is slightly more confusing, but, I've not worked with these types of documents before so deferring to you, Todd. Thoughts?

<details>
<summary>ReviewerReady "no" edge targets Activity 1 with a label that promises something else</summary>

<br/>plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/references/process-flow.dot:39
<br/>Caught by: Adversarial panel (Hawk and Owl rated Important; Dove conceded label is wrong)<br/>

The edge ReviewerReady -> Architecture [label="no, next\nunfinished activity"] has two contradictions at once.

Wrong target. SKILL.md §Activity 5 is explicit: self-review failures are fixed in place — "If you find issues, fix them inline and move on." Self-review is "explicitly the last step" and the prose describes no loop backward. The diagram invents a backward loop that re-runs Activities 1–4 (re-invoking architecting-solutions, re-mapping layers, re-scanning in-flight work, re-walking cross-team impacts) after any self-review nit, directly contradicting "fix inline and move on."

Misleading label. "next unfinished activity" is resume-dispatch vocabulary that implies conditional routing, but the edge unconditionally targets Architecture regardless of which activity is unfinished. The label makes a promise the edge cannot keep — there is exactly one outbound "no" edge and it has one hardcoded destination.

Fix:
ReviewerReady -> SelfReview [label="no, fix inline"];
ReviewerReady -> Done       [label="yes"];

</details>

theMickster (changes requested):
⚠️ Adding new skills should be a minor version bump; not patch bump. We have been fairly consistent with that across plugins. Will you please bump `bitwarden-delivery-tools` to 1.5.0?

❓/💡 Maybe not a now thing, but going back-n-forth between `developing-breakdown-spec` and `developing-breakdown-plan` is a little confusing. Trying to draw the connections and review this one while going back to what's in it's predecessor/sibling was difficult 😵‍💫

Regardless, handful of things to consider refactoring/clarifying. Couple general questions/ideas. Lemme know if you decide to defer any of the changes until after we get tech leads using the new skills.

trmartin4:
> ❓/💡 Maybe not a now thing, but going back-n-forth between developing-breakdown-spec and developing-breakdown-plan is a little confusing. Trying to draw the connections and review this one while going back to what's in it's predecessor/sibling was difficult 😵‍💫

@theMickster Do you think it makes sense to combine spec + plan skills into one then?  They are closely related and build upon one another; "doing a breakdown" implies both.

I have addressed all other feedback, I believe.  Thank you for the thorough review!

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
trmartin4: Addressed in [`46feb4c` (this PR)](https://github.com/bitwarden/ai-plugins/pull/143/commits/46feb4c5deaee0bb04a3f8da5f47f8b738c52b50).

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
trmartin4: Addressed in [`46feb4c` (this PR)](https://github.com/bitwarden/ai-plugins/pull/143/commits/46feb4c5deaee0bb04a3f8da5f47f8b738c52b50).

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
trmartin4: Addressed in [`46feb4c` (this PR)](https://github.com/bitwarden/ai-plugins/pull/143/commits/46feb4c5deaee0bb04a3f8da5f47f8b738c52b50).

I think that perhaps this whole section may need to go under a separate H2 in the breakdown to make this less brittle; I can do that in a follow-up PR if you agree.

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
trmartin4: Yep, fixed with Addressed in [`46feb4c` (this PR)](https://github.com/bitwarden/ai-plugins/pull/143/commits/46feb4c5deaee0bb04a3f8da5f47f8b738c52b50).  Thank you for catching that, I read in raw markdown not rendered after the latest updates.

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/references/worked-example.md
trmartin4: I removed this in [`46feb4c` (this PR)](https://github.com/bitwarden/ai-plugins/pull/143/commits/46feb4c5deaee0bb04a3f8da5f47f8b738c52b50).

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/references/process-flow.dot
trmartin4: yep, I think this makes sense.  Addressed in [`46feb4c` (this PR)](https://github.com/bitwarden/ai-plugins/pull/143/commits/46feb4c5deaee0bb04a3f8da5f47f8b738c52b50).

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/references/process-flow.dot
trmartin4: Yep, makes sense. We were instructing Claude to fix issues as it goes and not go back to previous steps, but the diagram didn't reflect that.  Addressed in [`46feb4c` (this PR)](https://github.com/bitwarden/ai-plugins/pull/143/commits/46feb4c5deaee0bb04a3f8da5f47f8b738c52b50).

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
trmartin4: Addressed in [`46feb4c` (this PR)](https://github.com/bitwarden/ai-plugins/pull/143/commits/46feb4c5deaee0bb04a3f8da5f47f8b738c52b50).  I guided the skill to prompt the user to navigate to the root folder.

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
trmartin4: For #1, I think that perhaps that instruction is too prescriptive, as I re-read it.  It feels like if the breakdown owner deleted a section, we should trust them as the arbiter, and not flag it.

For #2, I verified that we're using the correct `breakdown.md` everywhere in this branch.

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
trmartin4: I added this to the Key Principles:

> Bind untrusted-derived values as literal shell arguments. When interpolating breakdown-derived values (file paths, module names, team folders, repo names) into shell commands, pass them as fixed-string positional arguments — e.g. grep -F -- "$NAME". Never splice them into a shell-evaluated command string.

Does that sound right to you?

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
trmartin4: Added our Atlassian tools in [PR feedback.](https://github.com/bitwarden/ai-plugins/pull/143/commits/46feb4c5deaee0bb04a3f8da5f47f8b738c52b50).

plugins/bitwarden-tech-lead/CHANGELOG.md
trmartin4: Done in [PR feedback.](https://github.com/bitwarden/ai-plugins/pull/143/commits/46feb4c5deaee0bb04a3f8da5f47f8b738c52b50).

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
theMickster: I think your gut is right. Let's see how it lands, but likely a followup.

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
theMickster: All good. I thought something was wrong on my end 😅

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
```diff
@@ -0,0 +1,158 @@
...
+description: Develop the Plan section of a Bitwarden Tech Breakdown after the Specification is filled — technical architecture, per-layer impact, in-flight collision scan, cross-team impact mapping, and self-review. Supports resumption against a partly-developed Plan. Triggers: "develop the plan", "draft the implementation plan", "map per-layer impact", "scan for in-flight work", "identify cross-team impacts", "continue planning", "plan the breakdown".
+allowed-tools: Skill, Read, Edit, Write, Bash, Glob, Grep, TaskCreate, AskUserQuestion, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__get_issue, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__get_issue_comments, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__get_issue_remote_links, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__search_issues, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__get_confluence_page, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__get_confluence_page_comments, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__search_confluence, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__search_confluence_cql
```
SaintPatrck: ⛏️ `AskUserQuestion` and `TaskCreate` don't need to be explicitly listed.

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
```diff
@@ -0,0 +1,158 @@
...
+# Developing the Plan
+
+## Overview
+
+Assist a Bitwarden engineer in developing the HOW a change will be built, anchored to the already-defined Specification section of the breakdown document. The skill iterates on a technical architecture with the user, walks the change against every part of our technical stack to surface impact, scans for in-flight work that could collide, identifies and characterizes every cross-team impact, and runs a final self-review pass against the breakdown template.
+
+<HARD-GATE>
+Prompt the user to switch to their workspace root: the folder containing their local clone of `tech-breakdowns/` alongside the other Bitwarden repos (`server/`, `clients/`, `sdk-internal/`, `ios/`, `android/`, etc.). The skill relies on traversing those siblings to scan in-flight work and resolve cross-team impact.
+
+Orientation within a breakdown is required. Ask the user which breakdown to work against. They can give a path, a Jira key, or a team/slug — use `Glob` under `tech-breakdowns/` to resolve to a real `breakdown.md`. If the user already named it earlier in the conversation, confirm the resolved path with `AskUserQuestion` before proceeding.
```
SaintPatrck: 🎨 Would be helpful to coerce this proactively when invoked by adding `argument-hints` and/or `arguments` in frontmatter. Then the document can be referenced with `$ARGUMENTS[n]` or `$name` within the Skill body. 

More on `argument-hints` and `arguments` can be found [here](https://code.claude.com/docs/en/skills#frontmatter-reference).

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
SaintPatrck: ⛏️ A little weary of the unrestricted Skill permission. Can we scope to the ones explicitly referenced? Any others it finds will still be usable it will just force acceptance by the user.

plugins/bitwarden-delivery-tools/skills/developing-breakdown-plan/SKILL.md
trmartin4: Thank you for the docs!  I think I've used that correctly with the changes in [`78090f1` (this PR)](https://github.com/bitwarden/ai-plugins/pull/143/commits/78090f1458a024f7191eafcb52108ebf5da60a67).  I wasn't aware this was something you could do!

theMickster (approved):
I'm am good with this round of changes. Anything else at the moment feels too ⛏️-esk

PR #159, opened 2026-07-09
improve committing changes skill

quexten:
## 📔 Objective

Currently, the committing  changes skill may result in hard to review changesets. We want the change sets to be easily reviewable. To do this, they have to be structured by intent, with one intent captured per commit. This makes PR's easy to review commit-by-commit instead of as one big diff.

Example: https://github.com/bitwarden/sdk-internal/pull/1226/changes/406cfdc4e2f1c0961f69503e8ca18baa72e22230


plugins/bitwarden-delivery-tools/skills/committing-changes/SKILL.md
```diff
@@ -47,6 +52,18 @@ Update error handling in login flow
...
+
+Changes should be structured by intent. If there is more than one intent in the working directory / staged, break down the changes into a set of commits — one intent per commit (a refactor, a new helper, a feature, a bugfix, a test addition, a rename), each independently reviewable.
+
+---
+
+## Follow-up Fixes on an Open PR
+
+When committing a follow-up fix on a PR that is not yet merged, consider doing a rebase so that the single intent commit that introduced this change gets replaced, rather than stacking a separate "fix" commit on top. This keeps history structured by intent (see above): the branch reads as one clean commit per intent instead of a change plus a trail of corrections.
+
+---
```
SaintPatrck, line 65: 💭 I feel like `restructuring-git-history` warrants its own skill that can be triggered by users when/if they want to cleanup their work history. That skill would reference `committing-changes` to format a specific commit correctly. As a separate skill, it also becomes flexible enough to incorporate into other things like [this "Storybook" concept](https://github.com/bitwarden/ai-plugins/pull/111) I was experimenting with to improve the code review experience.

It's also important to consider the description and expected triggers for this skill. They're intentionally broad enough to trigger on any allusion to committing changes which increases chances of unintentionally and silently rewriting someone's history.

plugins/bitwarden-delivery-tools/skills/committing-changes/SKILL.md
```diff
@@ -25,6 +27,9 @@ description: Git commit conventions and workflow for Bitwarden repositories. Use
 
 Users reported confusion about when biometric prompts appear.
 This adds a configurable timeout setting to the security preferences.
+
+Jira-Ticket: PM-12345
```
SaintPatrck, line 31: ⚠️ This example now repeats the Jira ticket already included in the commit title. Is this meant to be an example of a commit addressing multiple tickets?

SaintPatrck (commented):
@quexten we spoke briefly in Slack about these changes possibly not being needed now that we have Stacked PR's enabled. Are you still interested in exploring this, or should it be closed for now?

❓ Is https://github.com/bitwarden/sdk-internal/commit/406cfdc4e2f1c0961f69503e8ca18baa72e22230 meant to be an example of a good or bad commit?

PR #160, opened 2026-07-10
Move architecting-solutions from bitwarden-tech-lead to bitwarden-delivery-tools

trmartin4:
## 📔 Objective

Moves the `architecting-solutions` skill from `bitwarden-tech-lead` into `bitwarden-delivery-tools` so it can be invoked directly from delivery-lifecycle skills (breakdowns, initiative-funnel work, ownership transitions) without requiring users to install `bitwarden-tech-lead`. This also mirrors the shared organizational responsibility of designing solutions - it is not isolated to a tech lead role.

As written previously, it would require anyone to install the tech-lead agent, which - given that all engineers are not tech leads and solution design is not a tech lead responsibility - presents a dissonance that would inhibit its invocation when potentially useful.

Removes references to the initiative shepherd, as this is a human-to-human process engaged outside of the solution architecture process, but kept a callout to engage with Architecture if we identify solution changes that warrant it.


plugins/bitwarden-delivery-tools/skills/architecting-solutions/SKILL.md
```diff
@@ -1,14 +1,16 @@
...
 allowed-tools: Skill, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__get_issue, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__get_issue_comments, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__get_issue_remote_links, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__search_issues, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__get_confluence_page, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__get_confluence_page_comments, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__search_confluence, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__search_confluence_cql
 ---
 
+This skill applies architectural judgment during team-level delivery work: scoping tickets, drafting breakdowns, evaluating trade-offs, and choosing between approaches.
```
SaintPatrck: ❓ How did adding this statement change Claude's behavior?

plugins/bitwarden-delivery-tools/skills/architecting-solutions/SKILL.md
```diff
@@ -44,35 +46,13 @@ Bitwarden is a password manager — security isn't a feature, it's the product.
 
 Teams have autonomy over decisions inside their domain. Architecture doesn't gate-keep team-level work. What Architecture does is maintain the holistic view — the portfolio of cross-cutting initiatives, the patterns that span teams, the decisions that will be expensive to change later. The job at the team level is to recognize when a choice has implications that benefit from that wider view, and pull Architecture in before — not after — the team ships.
 
-Watch for the five signals that warrant Architecture involvement (from the [Architecture / Engineering Operating Model](https://bitwarden.atlassian.net/wiki/spaces/EN/pages/1286963201)):
+Watch for signals that warrant Architecture involvement (from the [Architecture / Engineering Operating Model](https://bitwarden.atlassian.net/wiki/spaces/EN/pages/1286963201)):
```
SaintPatrck: ❓ Is Claude actually fetching this URL at any point, without explicit instruction to do so? Should it, so it is aware of the full decision model, or are we comfortable relying solely on the key points below? Is there concern with more drift between them? Does behavior degrade if this link is not present?

SaintPatrck (dismissed):
Left a couple of questions that are byproducts of not having evals associated with any of this. I don't want to hold the PR, but there would be real value in adding evals in a follow-up. Without them, reviewers have no way to verify if changes are improving or regressing behavior, and no way to identify whether we can remove excess commentary. Like submitting code changes without unit tests. We trust, but we also want to verify.

plugins/bitwarden-delivery-tools/skills/architecting-solutions/SKILL.md
trmartin4: You're right.  It doesn't meaningfully change it, it's not addressed to Claude but to a human.  I removed it.

plugins/bitwarden-delivery-tools/skills/architecting-solutions/SKILL.md
trmartin4: I removed the URL as it wasn't serving any meaningful purpose.

plugins/bitwarden-tech-lead/CHANGELOG.md
```diff
@@ -5,6 +5,17 @@ All notable changes to the `bitwarden-tech-lead` plugin will be documented in th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.0.0] - 2026-07-10
+
+### Removed
+
+- **BREAKING:** `architecting-solutions` skill moved to `bitwarden-delivery-tools` 2.1.0 so it can be invoked independent from a `bitwarden-tech-lead` agent. The skill is still addressable by name (`Skill(architecting-solutions)`) as long as `bitwarden-delivery-tools` 2.1.0+ is installed — which the tech-lead agent already lists as a required cross-plugin dependency.
```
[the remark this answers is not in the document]
SaintPatrck: @trmartin4 this is a valid finding.

plugins/bitwarden-delivery-tools/skills/architecting-solutions/evals/behavior-eval.json
SaintPatrck, line 1: 👍 This is a good start. We'll also want to track the baseline file (result of the current skill state) so we have data to compare against when re-running the eval.

plugins/bitwarden-tech-lead/CHANGELOG.md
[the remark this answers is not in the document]
trmartin4: Fixed in 728e71d. Both the tech-lead 3.0.0 and shepherd 1.0.1 changelog entries now reference `bitwarden-delivery-tools` 2.2.0.

plugins/bitwarden-delivery-tools/skills/architecting-solutions/evals/behavior-eval.json
trmartin4: @SaintPatrck how would we do that?  Is that committing versions of the skill along with the evals?

SaintPatrck: No need to maintain multiple versions of the skill. We only need to record and track the latest run's results in a `behavior-baseline.json` file. The expectation will be that evals are run and `behavior-baseline.json` updated any time changes are made. 

I haven't spent a ton of time dogfooding the baseline shape I'm currently using for behavior evals (below), so I don't want to be hyper-prescriptive about it yet, but as long as we can track, at a minimum, the `model + effort` the eval ran against and the pass/fail rate with and without the skill, then we can start quantifying the impact of changes.

```json
{
  "claude-opus-5/high": {
    "adr-applies-greenfield": { "skill": "5/5", "no-skill": "1/5" },
    "adr-deprecated-follow-superseder": { "skill": "5/5", "no-skill": "2/5" },
    "security-context-invocation-any-design": { "skill": "5/5", "no-skill": "0/5" },
    "security-fail-closed": { "skill": "4/5", "no-skill": "4/5" }
  }
}
```

With ☝️ we can quickly see that `adr-applies-greenfield` directives outperform the no-skill variant and diffing against the prior run we'd be able to see if the "with-skill" version improved or regressed after changes. After 2+ runs, if a grade goes from 4/5 -> 5/5 the new instruction improves behavior, if it goes from 5/5 -> 4/5 we know the instruction was detrimental. The no-skill variants signal if the skill/instruction is even needed with a specific model+effort combination. I.e., if `skill` and `no-skill` results are consistently equal, we can deduce that the skill/instruction is not necessary. We would only need to re-run the `no-skill` eval cases when a core model changes.

trmartin4: ah, ok.  So this `behavior-baseline.json` structure is something we came up with?  If so, I'd need to prompt Claude to use it?

SaintPatrck: Correct. I've not come across a standardized format, yet.

SaintPatrck: Looks like the `/skill-creator` Skill has a baseline schema defined for behavior evals. It looks fairly similar to what I've been using. Feel free to let `/skill-creator` drive the shape, since we're using it to scaffold and run the evals. They will probably align better. The shape I suggested is just what I've landed on through a little trial and error.

From `/skill-creator/references/schemas.md`
```json
"run_summary": {
  "with_skill":    { "pass_rate": {"mean": 0.85, "stddev": 0.05, "min": 0.80, "max": 0.90} },
  "without_skill": { "pass_rate": {"mean": 0.35, "stddev": 0.08, "min": 0.28, "max": 0.45} },
  "delta": { "pass_rate": "+0.50", "time_seconds": "+13.0", "tokens": "+1700" }
}
```

It's not immediately obvious how that ties back to each of your cases, but maybe `/skill-creator` will link them somehow when pointed to the actual eval. 🤷 

Side note, I like that it's tracking "tokens". Gives us a way to measure cost change. 🤔

trmartin4: @SaintPatrck I've added a baseline and `README` in [`9aa7fde` (this PR)](https://github.com/bitwarden/ai-plugins/pull/160/changes/9aa7fdee1bd79609a45fa78e452861aa36b53e3c).  I also restructured the evals to focus on the Bitwarden-specific parts of the skill, as that is the signal I feel like we want to make sure persists.

plugins/bitwarden-delivery-tools/skills/architecting-solutions/SKILL.md
```diff
@@ -0,0 +1,83 @@
...
+1. **WebFetch the ADR index** at `https://contributing.bitwarden.com/architecture/adr/`. Read every title. The corpus is small enough to scan in one pass.
+2. **Match every concern in your design against the corpus.**
+3. **Fetch each candidate ADR's page** and read the decision. Treat it as a constraint. If the ADR is marked Deprecated or Superseded, follow the superseder instead.
+
+### The ADR reference is the artifact that proves the check happened
+
+Every design you deliver must include a short **ADR reference** section that names:
+
+- Every ADR you consulted by name, and how it applies to your design.
+- Or, if no ADR governs the concerns in play, an explicit statement to that effect **after** actually scanning the index.
```
SaintPatrck, line 32: ♻️ More of a note to myself; this is a good candidate to be replaced with `consulting-adrs` Skill, once it's published.

SaintPatrck (approved):
🔥 
:shipit:

PR #163, opened 2026-07-13
Improvements to H1 Audit skill

mandreko-bitwarden:
## 📔 Objective

This inverts the direction of traversal in the `auditing-hackerone-vulns` skill so HackerOne is the source of truth instead of Jira, and folds in a batch of correctness fixes I've been running in my working copy.

The audit used to start from a `project = VULN` JQL sweep, which meant any open HackerOne report that never got a VULN ticket created was invisible to it. It now enumerates open reports across both programs (`bitwarden` VDP and `bitwarden-bbp` Bug Bounty) via the HackerOne MCP and correlates outward to VULN tickets, child engineering items, fix PRs, and release state.

### New action tokens

- 🆕 **Create VULN Ticket** for an open report with no linked VULN
- 🟣 **Close HackerOne Report** for a VULN already Closed or Rejected while the report stayed open
- 🏁 **Close Out** for a Verified VULN whose fix is confirmed in production but whose ticket was never moved to Closed

A reconciliation sweep and a 🗂️ Orphaned VULNs section catch open Jira tickets whose report was already closed, which a pure H1-first traversal would otherwise miss.

### Status handling

Child statuses gain a **Blocked** category, so `On Hold` work reports as stalled rather than collapsing into ⚪ Waiting, plus `Ready for Dev` and `In QA`. The query exclusion list widens to `(Done, Closed, Rejected, Resolved, Canceled)`, because VULN-project `status not in (Done, Verified)` does not actually exclude `Rejected` or `Closed`.

### Correctness fixes on the GitHub side

The PR search `--jq` filter now reads `.pull_request.merged_at`. The Search Issues API returns no `mergedAt` field, so the old filter reported every PR as unmerged. Release inclusion picks the earliest release whose `publishedAt` postdates the merge and confirms it by grepping the tag-to-tag commit range for the PR number, so a missing cherry-pick reports as 🔵 Monitor with an engineering flag rather than counting as shipped.

### Documented API constraints

These are all things I hit in practice and kept re-learning:

- Report IDs are base64 GIDs that need decoding
- The Jira reference lives in `get_report_activities`, not `get_report`
- `search_issues` silently drops the `issuelinks` field, so `linkedIssues()` is required
- `search_reports` requires `program_handles` and returns HTTP 500 for the `open` and `needs-more-info` states
- Jira's `Fix Version` "(Released)" annotation is untrustworthy

Also adds tool-usage guardrails against interpreter pipes, file writes, heredocs, shell orchestration, and `2>/dev/null` error suppression.

### Note on history

I force-pushed over the previous two commits on this branch. They had bumped the plugin to 1.3.0, which main then released independently as an unrelated ADR-alignment change, so the branch was conflicting. This version sits at 1.4.0 on top of current main and the changelog is sequential again.


plugins/bitwarden-security-engineer/skills/auditing-hackerone-vulns/SKILL.md
```diff
@@ -1,7 +1,7 @@
 ---
 name: auditing-hackerone-vulns
 description: Audit all open HackerOne-sourced VULN Jira tickets and their linked engineering child items to identify what needs action. Use this skill whenever the user wants to: check VULN ticket status, see which HackerOne findings need status updates, identify vulnerabilities ready to verify or close, run a remediation audit, check "what do I need to do on my VULN tickets today", or get a prioritized view of open vulnerabilities. Outputs a sorted action table with emoji tokens. Always use this skill for HackerOne/VULN remediation tracking and status correlation tasks — don't try to do it from scratch.
-allowed-tools: mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__search_issues, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__get_issue, mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__get_issue_remote_links, Bash(gh api --method GET *), Bash(gh pr view *), Bash(gh release list *), Bash(gh api repos/bitwarden/*/compare/*), Bash(gh search prs *)
```
withinfocus, line 4: ❓ Intentional?

plugins/bitwarden-security-engineer/skills/auditing-hackerone-vulns/SKILL.md
```diff
@@ -11,109 +11,153 @@ allowed-tools: mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__search
 | 🔴    | **Update VULN Status** | Child item has progressed (In Progress/Review) but VULN is still at a lower status                    |
 | 🟡    | **Mark Remediated**    | Child item is Done — set Remediation Date to merged PR date and move VULN to Remediated               |
 | 🟢    | **Verify & Close**     | Fix is in a release that has already shipped — verify in prod, add Confirmation Date, close HackerOne |
```
withinfocus: 🎨 As we discussed in chat, the confirmation date usually gets set shortly after one is remediated, so you know when the release will go out and therefore the validation.

plugins/bitwarden-security-engineer/skills/auditing-hackerone-vulns/SKILL.md
```diff
@@ -11,109 +11,153 @@ allowed-tools: mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__search
...
 
 ---
 
+## Tool usage rules — read before executing any step
+
+- **Only use tools listed in `allowed-tools`.** Do not use `curl`, `wget`, `python3`, `node`, or any interpreter or HTTP client not in the list.
+- **Do not write files.** Do not write to `/tmp/`, to any file path, or use heredocs (`cat > FILE << 'EOF'`). Hold all data in-memory between steps.
+- **Do not use shell parallelism.** Do not use `&` background processes, `wait`, `declare -A` / `typeset -A` associative arrays, or multi-step bash/zsh scripts. When this skill says "in parallel", it means issue multiple **Claude tool calls** in the same response turn — not shell concurrency.
+- **No interpreter pipes.** Do not pipe command output to `python3`, `node`, `ruby`, or any interpreter. Use `--jq` or standalone `jq` for JSON.
+- **No error suppression.** Do not add `2>/dev/null` to any command. Silent failures are indistinguishable from empty results and waste follow-up calls.
```
withinfocus, line 26: ❓ Did these cause issues? These are more general concerns of mine, because it's not skill-specific.

plugins/bitwarden-security-engineer/skills/auditing-hackerone-vulns/SKILL.md
```diff
@@ -11,109 +11,153 @@ allowed-tools: mcp__plugin_bitwarden-atlassian-tools_bitwarden-atlassian__search
...
+
+**Two-attempt cap on PR searches.** If the first search (by child item key) returns no results, try one more search using the VULN key (e.g., `VULN-529+type:pr+org:bitwarden`). If that also returns no results, mark "No PR found" and move on. Do not try keyword searches, repo-scoped retries, or other fallback queries — they rarely succeed and waste significant time.
+
+**Skip `gh pr view` when merge date is already known.** Only call `gh pr view PR_URL --json state,mergedAt,baseRefName,title` if the search returned `mergedAt: null`. If the search already returned a non-null merge date, you have what you need — do not make a redundant follow-up call. If `gh pr view` also returns `mergedAt: null`, the PR was not merged — record it as closed-without-merge and move on; do not search further.
+
+**Release list caching** — Fetch releases **once per repo** and reuse the result for all PRs in that repo. Do not re-fetch for each PR.
 
 ```bash
-gh pr view PR_URL --json state,mergedAt,mergeCommit,baseRefName,title
+# Note: gh release list is blocked — use the Releases API directly
```
withinfocus: ❓ Per my other question, why? Did the above change actually cause this?

withinfocus (commented):
Still hoping to see the prompting for dates when a code change happens. Automation in Jira can be added for some of that to actually prompt someone, but if you are already querying GitHub I would think this can call out what to set a confirmation date to.

plugins/bitwarden-security-engineer/skills/auditing-hackerone-vulns/SKILL.md
mandreko-bitwarden: These typically came as part of trying to get it so that the skill didn't have to have me accept a bunch of prompts. Things like interpreter pipes require a user to inspect the command because claude is unsure if it's dangerous or not. If we can avoid writing commands in those ways, without requiring user prompts, but still getting the same result, it's a win. I could see this type of thing bubbling up to a more global set of our skills or rules, but some may not apply.

plugins/bitwarden-security-engineer/skills/auditing-hackerone-vulns/SKILL.md
mandreko-bitwarden: the `gh release list` was conflicting with our rule here https://github.com/bitwarden/ai-plugins/blob/ccc5b2874862436674bf2e0fe43f3d1382932f14/plugins/bitwarden-code-review/.claude/settings.json#L29 where we deny the use of `gh release` as a potential security risk. I had talked to @theMickster about it at one point, and the issue comes in that it was getting too hard to allow certain safe usages while denying all the potentially dangerous ones, so it was easier to just deny all.

plugins/bitwarden-security-engineer/skills/auditing-hackerone-vulns/SKILL.md
mandreko-bitwarden: Answered above. `gh release` being blocked in other related skills causing conflicts.

plugins/bitwarden-security-engineer/skills/auditing-hackerone-vulns/SKILL.md
withinfocus: @SaintPatrck and team: I am of the position that we still don't want to include these sorts of instructions. These are elements of the run mode that a user takes on and something like Claude's auto mode won't have issues doing this.

plugins/bitwarden-security-engineer/skills/auditing-hackerone-vulns/SKILL.md
withinfocus: I would fix the other side. You are just trying to use a listing operation ...

plugins/bitwarden-security-engineer/skills/auditing-hackerone-vulns/SKILL.md
mandreko-bitwarden: I keep trying to figure out a good way to calculate the confirmation date, and haven't come up with anything. We have to take the merge date of a PR, compare it to a release cut date, and then decide which release it will be in accordingly. But getting the release calendar into this skill is a bit of a mess each time I try. This is one of those things that may be easier to handle manually?

Do you have any thoughts on how to handle this better?

mandreko-bitwarden:
> * the `auditing-hackerone-vulns` skill is not listed in the agent's `skills:` block

This is definitely on purpose. This is a stand-alone skill

plugins/bitwarden-security-engineer/skills/auditing-hackerone-vulns/SKILL.md
withinfocus: You should just do something like prompt the user to enter the confirmation date.

withinfocus (changes requested):
See the validator comment -- those majors should be resolved. The skill should be broken out to references and probably thinned into several skills; you have a lot of special cases through this and it can't all be in one skill file for this to operate effectively. "Auditing" these is really a large request that needs composition.

Claude loves to make giant changelogs too, so once the above is done make sure that's compressed to meet our norms.

mandreko-bitwarden:
> See the validator comment -- those majors should be resolved. The skill should be broken out to references and probably thinned into several skills; you have a lot of special cases through this and it can't all be in one skill file for this to operate effectively. "Auditing" these is really a large request that needs composition.
> 
> Claude loves to make giant changelogs too, so once the above is done make sure that's compressed to meet our norms.

Yea, I'm going to move this back to Draft for now. I was working on it last night, and seeing similar suggestions too. It still needs additional work at this point and shouldn't be poised as ready for review. 

[EDIT: Doh! It is already in Draft..]

PR #164, opened 2026-07-14
[AI-2] feat: add bitwarden-ai-telemetry plugin for metadata-only AI-usage telemetry

SaintPatrck:
## 🎟️ Tracking

[AI-2](https://bitwarden.atlassian.net/browse/AI-2)

## 📔 Objective

Adds the `bitwarden-ai-telemetry` plugin: Claude Code hooks that emit metadata-only AI-usage telemetry as OTLP logs. Native Claude Code telemetry redacts or omits several identifiers that matter for understanding how AI-assisted work actually happens — skill and agent names collapse to generic placeholders like `"custom"` and `"custom_skill"`, and MCP tool calls collapse to `"mcp_tool"`. This plugin's hooks recover the real names and also capture the git-linkage between a session and the repo, branch, files, commits, and PRs it touched.

Everything captured is metadata: names, paths, SHAs, PR numbers, and session IDs. Prompts, file contents, and tool arguments are never captured. Every hook fails open, so a misconfigured or unreachable destination never blocks a session.

`emit.py` has no default collector URL. It only sends telemetry once `BW_TELEMETRY_OTLP` is explicitly set, normally through managed-settings.json's `env` block.

Test coverage now exists for `emit.py`, `emit_identity.py`, and `emit_mcp.py`, extracted into pure functions the same way `emit_git.py` already was.

[AI-2]: https://bitwarden.atlassian.net/browse/AI-2?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/bw-telemetry/README.md
```diff
@@ -0,0 +1,62 @@
...
+Telemetry is best-effort and must never interfere with a working session. Every hook is guarded: the shared emitter swallows all errors, git and network calls are time-boxed, and each hook always exits `0`. If the destination is unreachable, misconfigured, or slow, the session proceeds unaffected and the record is simply dropped.
+
+## Configuration
+
+The OTLP destination endpoint is **not hardcoded** — it is supplied at deploy time via environment variables:
+
+- `BW_TELEMETRY_OTLP` — the full OTLP logs endpoint URL the hooks POST to.
+- Standard `OTEL_*` environment variables for OpenTelemetry configuration.
+
+When `BW_TELEMETRY_OTLP` is unset, the emitter falls back to a local loopback collector (`http://127.0.0.1:4318/v1/logs`), so nothing is sent off-machine unless an endpoint is explicitly configured.
```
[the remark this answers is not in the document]
SaintPatrck: @claude resolved in [3ce5fb7](https://github.com/bitwarden/ai-plugins/commit/3ce5fb7849f03f186ef12f6cb87e8cc5f8c2e001)

plugins/bw-telemetry/hooks/emit_git.py
```diff
@@ -0,0 +1,178 @@
...
+        m = re.search(r"/pull/(\d+)", out) or re.search(r"#(\d+)\b", out)
+        if m:
+            emit("bw.pr", {
+                "event.name": "bw.pr",
+                "session.id": session,
+                "bw.repo_full": _repo_full(cwd),
+                "bw.branch": _git(cwd, "rev-parse", "--abbrev-ref", "HEAD"),
+                "bw.pr_number": m.group(1),
+                "bw.hook": h.get("hook_event_name", ""),
+            })
```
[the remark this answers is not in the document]
SaintPatrck: @claude resolved in [be5f84f](https://github.com/bitwarden/ai-plugins/commit/be5f84f04d5ed0a6a8fb8635a7c01b2d9289cecf)

prograhamming (dismissed):

plugins/bitwarden-ai-telemetry/hooks/emit.py
```diff
@@ -0,0 +1,37 @@
...
+    payload = {"resourceLogs": [{
+        "resource": {"attributes": [
+            {"key": "service.name", "value": {"stringValue": "bitwarden-ai-telemetry"}}]},
+        "scopeLogs": [{"scope": {"name": "bw.telemetry.hooks"},
+                       "logRecords": [{"timeUnixNano": str(time.time_ns()),
+                                       "body": {"stringValue": body_name},
+                                       "attributes": kv}]}]}]}
+    try:
+        req = urllib.request.Request(
+            COLLECTOR, data=json.dumps(payload).encode(),
```
withinfocus, line 59: `COLLECTOR` goes straight into `urllib.request` with no scheme check. I confirmed locally that urllib honors both `http://` (plaintext) and `file://` here. If the collector is ever misconfigured as `http://`, the repo slug, branch names, file paths, and session ids cross the network in the clear, which breaks the secure-channel guarantee (SC). Suggest rejecting anything that isn't https before the POST, e.g. `if urlsplit(COLLECTOR).scheme != "https": return`.

plugins/bitwarden-ai-telemetry/hooks/emit.py
```diff
@@ -0,0 +1,37 @@
...
+            headers={"Content-Type": "application/json"})
+        urllib.request.urlopen(req, timeout=1).read()
```
withinfocus: The response body is never used, but `.read()` pulls all of it into memory from a destination we don't control. It's bounded by the 1s socket timeout and the 5s hook timeout, so the risk is small. You can drop the `.read()` entirely.

plugins/bitwarden-ai-telemetry/README.md
```diff
@@ -0,0 +1,59 @@
...
+- Tool arguments or tool results
+- Any credentials or secrets
+
+## Fail-open by design
+
+Telemetry is best-effort and must never interfere with a working session. Every hook is guarded: the shared emitter swallows all errors, git and network calls are time-boxed, and each hook always exits `0`. If the destination is unreachable, misconfigured, or slow, the session proceeds unaffected and the record is dropped.
+
+## Configuration
+
+The OTLP destination is supplied at deploy time via the `BW_TELEMETRY_OTLP` environment variable (normally set org-wide through managed-settings.json's `env` block), and is not hardcoded anywhere in the plugin.
```
withinfocus, line 41: "Normally set org-wide through managed-settings.json" is the soft spot. Managed settings only take precedence where a managed policy exists. On a machine without one, a project-level `.claude/settings.json` env block shipped inside a cloned repo can set `BW_TELEMETRY_OTLP` and redirect that developer's session metadata (repos, branches, file paths, session ids, which internal skills/agents/MCP servers they run) to an endpoint the repo author chose. The fail-open design means nothing surfaces to the user when that happens. Worth either pinning the host to an allowlisted Bitwarden domain in `emit.py`, or stating here that this variable must be set only via managed settings and must not be honored from project settings.

plugins/bitwarden-ai-telemetry/README.md
```diff
@@ -0,0 +1,59 @@
...
+- Commit SHAs and PR numbers
+- Tool, skill, agent, and MCP server/tool **names**
+- The Claude Code session id
+
+**It never collects:**
+
+- File contents or diffs
+- Prompt text or model responses
+- Tool arguments or tool results
+- Any credentials or secrets
```
withinfocus: The code can't actually guarantee this. Branch names and file paths are emitted verbatim, and those occasionally carry a token or an internal codename. Suggest softening to something like "does not intentionally collect credentials; branch and file names are captured as-is" so the claim matches what the hooks can enforce.

plugins/bitwarden-ai-telemetry/hooks/emit_git.py
```diff
@@ -0,0 +1,211 @@
...
+        return False
+    # Require a positive stdout signal, same as _is_successful_commit, so an
+    # empty/ambiguous response never fabricates a PR.
+    return bool(stdout)
+
+
+def _git(cwd, *args):
+    """Run a git command in cwd, returning trimmed stdout or "" on any failure."""
+    try:
+        out = subprocess.run(["git", "-C", cwd, *args],
```
withinfocus, line 108: Worth noting these git calls now run automatically on every edit and Bash tool use, in whatever directory the session sits in, including untrusted cloned repos. `remote get-url` and `rev-parse` don't trigger git's config-driven execution paths (aliases, pager, fsmonitor), so this isn't an execution vector today. Flagging it so the git surface stays limited to these read-only plumbing commands and doesn't grow into subcommands that honor repo config.

withinfocus (commented):
Security audit of the telemetry hooks. The design is careful overall: argv-list subprocess with no `shell=True`, metadata-only emits, fail-open everywhere, no default collector, and cert verification on https. Findings below, most are low; the https-enforcement one is the only thing I'd want before merge. I verified the no-ReDoS and urllib scheme behavior locally rather than assuming.

plugins/bitwarden-ai-telemetry/README.md
SaintPatrck: 💭 I wanted to avoid hardcoding any specific endpoint, primarily to avoid exposing internal infra. Thoughts on limiting to `*.bitwarden.pw`?

withinfocus: I was thinking something like that, yeah.

withinfocus (approved):
Great improvements, thanks!

prograhamming (approved):

SaintPatrck:
Thanks @withinfocus and @prograhamming

PR #168, opened 2026-07-21
llm: align security assessments with Bitwarden ADRs

sognefej:
## 🎟️ Tracking

No Jira ticket — this work isn't tied to a tracked issue.

**What prompted this:** these skills weren't explicitly checking proposals or assessments against existing Architecture Decision Records, so they had no way to distinguish "this contradicts a decision we already made" from "this is a new architectural gap that needs a decision." Both cases were being treated the same way (or not flagged at all), which meant an assessment could satisfy P01-P06 in the abstract while still silently re-deciding or contradicting something already accepted.

## 📔 Objective

Security assessments could satisfy P01-P06 in the abstract while still contradicting an already-accepted architecture decision. `bitwarden-security-context`, `reviewing-security-architecture`, and `threat-modeling` now check Bitwarden's [Architecture Decision Records](https://contributing.bitwarden.com/architecture/adr/) for relevant prior decisions:

- Conflict with an accepted ADR is treated as a finding, not a style preference.
- A significant undocumented architectural decision is flagged as a gap.
- ADR status (accepted vs. superseded/deprecated) is verified before being cited.

Also includes the required version bump (1.2.0 → 1.3.0, MINOR — backward-compatible addition), changelog entry, and a `.cspell.json` addition for `ADRs`. `validate-plugin-structure.sh` and `validate-marketplace.sh` both pass.


plugins/bitwarden-security-engineer/skills/threat-modeling/SKILL.md
```diff
@@ -9,10 +9,11 @@ Bitwarden follows a 4-phase engagement model for security work. This skill prima
 
 ### Phase 1: Initial Security Assessment (Engineering Team)
 
-1. Create data flow diagrams (Mermaid, Excalidraw, or Structurizr)
-2. Define security requirements separate from product requirements
-3. Propose security definitions (threat model + security goals)
-4. Identify initial threats using STRIDE (see `references/stride-framework.md`)
+1. Check Bitwarden's [Architecture Decision Records](https://contributing.bitwarden.com/architecture/adr/) for existing decisions covering the system or components under review — a threat model should align with already-accepted architecture, not silently re-derive or contradict it
+2. Create data flow diagrams (Mermaid, Excalidraw, or Structurizr)
```
withinfocus: 🎨 Since you're here, remove everything but Mermaid. That's an ADR now 😜

withinfocus (changes requested):
I sent a DM about running the install steps too, but this is pretty close to ready; one comment.

For the PR in the tracking section, can you also document what might have prompted this change?

plugins/bitwarden-security-engineer/skills/threat-modeling/SKILL.md
sognefej: Fair enough 😅

SaintPatrck:
Apologies for pushing 6f71cfe. That was meant to stay local while I reviewed. The new "click" support in Claude Code terminal UI is a trap. 😆 

6f71cfe represents a change I was going to suggest: consolidate the ADR consulting guidelines into a single `/references` file instead of repeating the same instructions in multiple places.

♻️ I've noticed ADRs being referenced and consulted in a number of skills, workflows, and agents. A potential future improvement could convert it into a `consulting-adrs` skill. A dedicate skill would standardize checks against ADRs and make it reusable in other plugins.

plugins/bitwarden-security-engineer/skills/threat-modeling/SKILL.md
sognefej: fixed in commit 58f9552201c6017386863888d2767f15f707b4ce

sognefej:
Addressed both items from review:
- PR description now explains what prompted this change (Tracking section).
- `threat-modeling/SKILL.md` diagram tooling trimmed to Mermaid only (`58f9552`).

@withinfocus mind taking another look?

withinfocus (approved):
Works for me!

SaintPatrck (approved):

PR #169, opened 2026-07-22
Add bitwarden-testing-tools plugin with assessing-test-coverage skill

nthompson-bitwarden:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/QA-1983

## 📔 Objective
Add the bitwarden-test-toolkit plugin to the marketplace, providing an assessing-test-coverage skill that analyzes a repository's existing test stack and reports coverage gaps — with supporting references and evals, plus the required marketplace registration and dictionary/README updates.


SaintPatrck:
💭 @kdenney @nthompson-bitwarden this looks like it would compliment #140 really well. I can foresee the two being combined into an "assess -> generate -> execute" workflow. In order to more easily facilitate that I suggest they be siblings in a more generically named `bitwarden-testing-tools` plugin. That would also establish a common home for future testing related tooling. Thoughts?

nthompson-bitwarden:
> 💭 @kdenney @nthompson-bitwarden this looks like it would compliment #140 really well. I can foresee the two being combined into an "assess -> generate -> execute" workflow. In order to more easily facilitate that I suggest they be siblings in a more generically named `bitwarden-testing-tools` plugin. That would also establishes a home for future testing related tooling. Thoughts?

I think there's definitely more to consider and plan re: agentic testing. 

One example - Playwright can't install extension on Firefox just yet. I've been pushing https://github.com/webdriverio/mcp which has proven to be able to install the extension to Firefox and that has support for Appium (not only for mobile, but apparently can automate macOS desktop native applications as well. That _could_ mean macOS Desktop + Safari extension)

There _is_ an argument for "why not both?" - that way we get the cross-platform advantages of WebdriverIO MCP, but the more chromium-focused benefits of Playwright CLI. 

In _any event_ I'm happy to rename the directory from "bitwarden-test-toolkit" to "bitwarden-testing-tools" to match up with the rest of the repo

SaintPatrck:
> I think there's definitely more to consider and plan re: agentic testing.

I agree, there's still a lot to flesh-out before we dive into fully agentic testing. These PRs are a great foundation to start from.

> One example - Playwright can't install extension on Firefox just yet. I've been pushing https://github.com/webdriverio/mcp which has proven to be able to install the extension to Firefox and that has support for Appium (not only for mobile, but apparently can automate macOS desktop native applications as well. That could mean macOS Desktop + Safari extension)
>
> There is an argument for "why not both?" - that way we get the cross-platform advantages of WebdriverIO MCP, but the more chromium-focused benefits of Playwright CLI.

I'm definitely on team "both". 😆 Claude recently gained native support for running macOS applications too. We can do some testing and figure what combination of tools will work best.

> In any event I'm happy to rename the directory from "bitwarden-test-toolkit" to "bitwarden-testing-tools" to match up with the rest of the repo

Yea, let's go ahead and rename it. @kdenney reached out and is in consensus.

SaintPatrck:
I think Claude is overreacting to some of [these](https://github.com/bitwarden/ai-plugins/pull/169#issuecomment-5048513526) findings. 

Number 1 is an issue in `researching-jira-tickets`, not with this skill. It has the proper tool scopes declared for what it's using. We will need to update the research skill.

Number 2 is the only one that really needs to be addressed before merging, imo. 

Number 3 is interesting. If the eval is passing, I have reservation about adding more guidance based on speculation. Let's wait until we see failures around this scenario before making changes. Mentioning branch in the argument-hints would be a nice addition, but not absolutely necessary.

Number 4 sounds legit, but more of a nitpick. We expect it to be installed, although it doesn't hurt to add a short escape hatch sentence; "If `bitwarden-atlassian-tools` is not installed, stop and prompt the user to install it before continuing."

For the minor findings, I think the only ones really worth acting on now are the "Eval harness" ones. Claude should be able to knock those out easily enough.

nthompson-bitwarden:
thanks @SaintPatrck, I've taken care of those as well as some resolving some eval query findings that were leftover from previous iterations

SaintPatrck (approved):
This is excellent! Looking forward to seeing it action. 🥳

PR #192, opened 2026-08-10
[BRE-1830] feat(bitwarden-devops-engineer): Add naming conventions

fntyler:
## 🎟️ Tracking

[BRE-1830](https://bitwarden.atlassian.net/browse/BRE-1830)

## 📔 Objective

Add a ~`workflow-naming-conventions`~ `auditing-workflow-conventions` reference skill covering the GitHub Actions naming standards that `bwwl` does not enforce, so audits stop relying on undocumented convention.

The skill defers to `bitwarden-workflow-linter-rules` for every linted rule to avoid double-reporting, and treats job ID and file renames as report-only because both are addressable identifiers.

* Document job IDs (kebab-case), step names (Sentence case with a leading imperative verb), and workflow file names (kebab-case.yml, `_` prefix when exclusively reusable)
* Add reference-sweep procedures for job ID and filename renames, plus an advisory canonical step-name glossary
* Bump plugin to 0.2.0 with changelog entry and README listing
* Add `greppable` to the cspell dictionary

[BRE-1830]: https://bitwarden.atlassian.net/browse/BRE-1830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/bitwarden-devops-engineer/skills/workflow-naming-conventions/SKILL.md
```diff
@@ -0,0 +1,122 @@
+---
+name: workflow-naming-conventions
```
fntyler: Any reason _not_ to include something like `user-invocable: false` for skill definitions of this style/type? [Frontmatter reference](https://code.claude.com/docs/en/skills#frontmatter-reference)

SaintPatrck: Generally speaking, no, but I'm curious how you envision it to be used if it's not. Will it be referenced in another skill or agent somewhere? Also curious why you wouldn't want a user to be able to invoke it.

plugins/bitwarden-devops-engineer/skills/workflow-naming-conventions/SKILL.md
```diff
@@ -0,0 +1,122 @@
...
+
+  <example>
+  User: Should this reusable workflow file be named build.yml or _build.yml?
+  Action: Consult this skill for the workflow-file-naming standard
+  </example>
+
+  <example>
+  User: Review these workflows for naming consistency
+  Action: Apply the standards in this skill; treat glossary deviations as advisory
+  </example>
```
SaintPatrck: ⚠️ Per Anthropics' [Skill best practices](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices#skill-structure), frontmatter should not contain xml.

♻️ Other skills in this plugin incorrectly use XML, and they will need to be fixed in a separate PR.

SaintPatrck (changes requested):
Consider having [Skill Creator](https://github.com/anthropics/claude-plugins-official/tree/main/plugins/skill-creator) generate evals to provide evidence that all the prose within the new Skill is actually load-bearing, and not overly verbose as Claude likes to do. Without evals, objective review of the Skill content is impossible unless reviewers run and evaluate it manually.

plugins/bitwarden-devops-engineer/skills/workflow-naming-conventions/SKILL.md
fntyler: I was unaware of `user-invocable` or `disable-model-invocation` controls for skills prior to working on this one. In this case, this skill is more of a reference (similar to `bitwarden-workflow-linter-rules`) which I think are better served if invoked by an agent or another skill.

plugins/bitwarden-devops-engineer/skills/workflow-naming-conventions/SKILL.md
fntyler: Thanks for feedback. Addressed in https://github.com/bitwarden/ai-plugins/pull/192/commits/92baa42183b9f0045fcab55ebb9b47e0f96038dc

plugins/bitwarden-devops-engineer/README.md
```diff
@@ -15,6 +15,7 @@ The skills work in pairs. Each audit skill is strictly read-only and produces a
 | `action-audit`                    | Searches org-wide for compromised, deprecated, or unpinned GitHub Actions and produces a findings report. Read-only. |
 | `action-remediate`                | Applies hash pins or action replacements across selected repos and creates draft PRs based on audit findings.        |
 | `bitwarden-workflow-linter-rules` | Reference for all 10 `bwwl` linter rules — triggers, fix procedures, and mechanical vs. judgment categorization.     |
+| `workflow-naming-conventions`     | Reference for naming standards `bwwl` does not enforce — job IDs, step names, and workflow file names.               |
```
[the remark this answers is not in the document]
fntyler: No. At least not yet. The current shape of the `workflow-audit` doesn't fit cleanly without making load-bearing changes to `workflow-audit`.

plugins/bitwarden-devops-engineer/skills/workflow-naming-conventions/SKILL.md
SaintPatrck: Makes sense. Thanks for explaining. I have no qualms with `user-invocable: false`. Easy enough to undo later, if needed.

plugins/bitwarden-devops-engineer/skills/workflow-naming-conventions/SKILL.md
```diff
@@ -0,0 +1,111 @@
...
+  advisory canonical step-name glossary and reference-sweep procedures for job ID and filename
+  renames. Use when auditing or authoring workflows and questions like "what casing should job IDs
+  use", "should this reusable workflow be build.yml or _build.yml", or "review these workflows for
+  naming consistency" come up. Read alongside bitwarden-workflow-linter-rules, which is the source
+  of truth for linted rules; this skill covers only the gaps.
+---
+
+## Ownership
+
+This skill covers only what `bwwl` cannot check. For anything the linter enforces, `${CLAUDE_PLUGIN_ROOT}/skills/bitwarden-workflow-linter-rules/SKILL.md` is the source of truth — do not report a finding here that duplicates a linter rule.
```
SaintPatrck: ⛏️ With the plugin installed, this skill will be registered in context. You should be able to reference it with `Skill(bitwarden-devops-engineer:bitwarden-workflow-linter-rules)` instead of the full path. 

ℹ️ It's important to have the Skill **invoked** instead of simply **read** in case there are scripts or other complimentary files packaged in that referenced skill. Moot here since that skill is only the one markdown file, but if that every changes we'll want to update this reference.

⛏️ Since you're deferring work to, and instructing Claude to invoke the skill, consider adding it to `allowed-tools` in frontmatter so it doesn't force an approval prompt if Claude tries to use it.

Not considering either of these blocking if the skill is already working as expected.

plugins/bitwarden-devops-engineer/skills/workflow-naming-conventions/SKILL.md
```diff
@@ -0,0 +1,111 @@
+---
+name: workflow-naming-conventions
```
SaintPatrck: 💭 Names using gerund-verb form help Agents decide when to invoke a skill and reinforce the `description` to ensure reliable and consistent triggering. What do you think about `auditing-workflow-conventions` as the Skill name? I suggest that because it aligns very well with the action an agent would be taking, and is backed by the description, "Use when auditing or authoring workflows..."

I see that `workflow-linter-rules` already deviates from that convention, so not blocking. If you do change this Skills name, I suggest updating the name of `bitwarden-workflow-linter-rules` in a follow-up.

SaintPatrck (dismissed):
Content looks good. Couple of minor notes about naming and sibling skill references. Nothing worth blocking on though.

plugins/bitwarden-devops-engineer/skills/workflow-naming-conventions/SKILL.md
fntyler: Good to know. I wasn't sure if there was a standard that we wanted to adhere to. I'll keep it in mind for future reference. I'll keep it omitted for now. Thanks for the feedback!

plugins/bitwarden-devops-engineer/skills/workflow-naming-conventions/SKILL.md
fntyler: The current structure and flow (audit -> remediate) of the existing skills in this plugin likely warrant some reshaping to better adopt the practices that evolved.

> What do you think about auditing-workflow-conventions as the Skill name?

Feels like a foot in the right direction. Reflected in https://github.com/bitwarden/ai-plugins/pull/192/commits/e0c2fcdd175c3dcfd2be562bf6231951af89102e

plugins/bitwarden-devops-engineer/skills/workflow-naming-conventions/SKILL.md
fntyler: Appreciate the time and feedback. Addressed in both: https://github.com/bitwarden/ai-plugins/pull/192/commits/da050de6a970a7b13d6978b9ab063d68806f43fe & https://github.com/bitwarden/ai-plugins/pull/192/commits/988ae3f46566e63781c7bb982663715e7db1ce18

SaintPatrck (approved):

PR #209, opened 2026-08-21
feat(bitwarden-testing-tools): add mailcatcher and Stripe skills for testing

kdenney:
## 🎟️ Tracking

No Jira ticket.

Foundation layer (1 of 5) of stacked prs adding a fully-automated Playwright testing pipeline. Stack, bottom to top: **#209** (this PR, base `main`), #210, #211, #212, #213. Each layer is independently reviewable; review and merge bottom-up.

## 📔 Objective

The foundation layer of the testing pipeline. It bumps `bitwarden-testing-tools` to 1.2.0 and adds the two agent-less "tool" skills that a person can invoke directly, plus the shared eval runner the plugin's other skills build on.

- `reading-mailcatcher-api`: reads Bitwarden emails from the local Mailcatcher REST API (verification links, magic links, tokens) through a co-located `read_mailcatcher.py`. Preferred over the Mailcatcher browser UI in automated contexts, where Playwright's browser CORS blocks a direct fetch. It reads messages only and does not start, stop, or health-check the Mailcatcher container.
- `using-stripe-cli`: queries read-only Stripe test-mode data through a `stripe_cli.py` wrapper, with one permitted write, advancing an already-attached test clock. The wrapper builds every command itself and never forwards a caller-supplied flag, so it cannot be steered to live or production data.
- `scripts/eval_harness.py`: a shared trigger-eval runner that each skill's `evals/run_real_eval.py` configures rather than copies. The existing `assessing-test-coverage` eval runner is refactored down onto it; that skill is otherwise untouched.

Each new skill ships with unit tests, references, and a trigger eval recorded as a dated on-demand prose reading against the current four-skill inventory rather than a committed baseline, to avoid repeated baseline runs with frequent edits, though they could be added once changes slow down. `using-stripe-cli` also carries advice-only behavior evals.


plugins/bitwarden-testing-tools/skills/reading-mailcatcher-api/scripts/read_mailcatcher.py
```diff
@@ -0,0 +1,319 @@
...
+    parser.add_argument("--pattern", default="")
+    parser.add_argument("--link-filter", default=DEFAULT_LINK_FILTER)
+    # --mailcatcher-url / --allowed-host let the base-URL and allowlist
+    # overrides ride inside the granted `script *` argv shape. A leading
+    # `MAILCATCHER_URL=... read_mailcatcher.py` env assignment would move the
+    # command off the script path and miss the allowed-tools grant, prompting
+    # every run; a flag does not. Both default to the env vars, so the flag
+    # wins when given, the env var otherwise, the built-in default last.
+    parser.add_argument("--mailcatcher-url", default=None)
+    parser.add_argument("--allowed-host", action="append", default=[], metavar="HOST")
```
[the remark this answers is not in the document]
kdenney: Fixed by removing the base-URL override entirely rather than shape-constraining hosts. The Mailcatcher endpoint is the same across our dev environments, so `read_mailcatcher.py` now hardcodes `http://localhost:1080` and no longer accepts `--mailcatcher-url`, `--allowed-host`, or the `MAILCATCHER_URL` env var. A granted call can only pass `--recipient`/`--pattern`/`--link-filter`, so it can never point the fetch at another host, and `check_base` is gone with no caller-supplied host left to validate.

The extracted-URL allowlist stays, still extendable via the operator-supplied `PLAYWRIGHT_TESTING_ALLOWED_HOSTS` env var (a real env assignment, not auto-approved argv). Added a regression test asserting a hostile `MAILCATCHER_URL` is ignored.

plugins/bitwarden-testing-tools/skills/reading-mailcatcher-api/SKILL.md
```diff
@@ -0,0 +1,58 @@
+---
+name: reading-mailcatcher-api
+description: This skill should be used when a local testing or debugging flow needs a link or token from an email, such as account verification links, magic-link logins, trial activations, organization invites, emergency access, and other email-driven flows. It reads an email from the local Bitwarden Mailcatcher inbox and extracts a URL or token, querying the Mailcatcher REST API at http://localhost:1080 by recipient and subject; preferred over the browser UI in automation because Playwright CORS blocks direct fetch. Do NOT use it to configure SMTP, start Mailcatcher, or debug delivery.
```
SaintPatrck, line 3: 🎨 Can we trim this by focusing on **what it does** instead of **how it works**? The preference over Mailcatcher UI is worth keeping, since it helps with routing, but some of the other details shouldn't have an impact on triggering since they focus on the implementation details.

```suggestion
description: Retrieves a link or token from Mailcatcher. Use when a local testing or debugging flow needs a link or token from an email, such as account verification links, magic-link logins, trial activations, organization invites, emergency access, and other email-driven flows. Preferred over the Mailcatcher browser UI in automation, where Playwright's CORS blocks a direct fetch. Do NOT use it to configure SMTP, start Mailcatcher, or debug delivery.
```

📓 If we do go with that change the trigger evals will need to be rerun.

plugins/bitwarden-testing-tools/skills/reading-mailcatcher-api/SKILL.md
```diff
@@ -0,0 +1,58 @@
...
+argument-hint: "--recipient <email> [--pattern <subject-keyword>] [--link-filter <regex>]"
+allowed-tools: "Read, Bash(${CLAUDE_PLUGIN_ROOT}/skills/reading-mailcatcher-api/scripts/read_mailcatcher.py:*), Bash(${CLAUDE_PLUGIN_ROOT}/skills/reading-mailcatcher-api/scripts/get_admin_email.py:*)"
+---
+
+# Reading the Mailcatcher API
+
+## Quick reference — use the script
+
+For all programmatic uses (test runs, ad-hoc fetches, debugging), call the co-located script directly. It is user-invocable: trigger it from any Claude Code session with the arguments below and Claude runs the script and returns the extracted URL (or the `NO_MATCH` diagnostic).
```
SaintPatrck, line 12: ♻️ Phrases like "from any Claude Code session..." and "Claude runs the script..." makes this read like user-facing documentation instead of LLM instruction. Can we either remove it if it's meant to be user-facing (already stated in `README`), or rephrase it if it is load-bearing for the agent?

plugins/bitwarden-testing-tools/skills/reading-mailcatcher-api/SKILL.md
```diff
@@ -0,0 +1,58 @@
...
+For example, to pull a verification link:
+
+```
+${CLAUDE_PLUGIN_ROOT}/skills/reading-mailcatcher-api/scripts/read_mailcatcher.py --recipient testuser-s1@example.com --pattern "Verify"
+```
+
+- **`--pattern`** is optional; omit it to select the most recent message for the recipient
+- **`--link-filter`** is optional; it defaults to `verify|confirm|signup|token|trial|login|finish-signup` — the starting filter that the exit-1 rows below may tell you to loosen or fix
+- **stdout** (on success): the extracted URL, ready to navigate to or paste into a form field
+- **exit code** (on failure): see the `read_mailcatcher.py` table below (the `get_admin_email.py` helper has its own codes, noted where it is introduced)
```
SaintPatrck: ❓ What is the parenthetical doing? Does Claude innately reach for `get_admin_email.py` without it?

plugins/bitwarden-testing-tools/skills/using-stripe-cli/SKILL.md
```diff
@@ -0,0 +1,116 @@
...
+- **Subscription issues:** check `status`, the expanded `latest_invoice`, and recent `customer.subscription.*` events.
+- **Event tracing:** `/v1/events` filters by type, not object ID; list by type and filter client-side. Events carry the object snapshot at event time.
+- **Customer state:** expand `subscriptions` and read `--path /v1/customers/<id>/payment_methods` for the full picture.
+
+## The one permitted write: advancing an already-attached test clock
+
+You may advance a test clock that is already attached to the subscription. You may NOT attach a clock to an existing subscription: the Stripe CLI cannot do it (a test clock can only be set on a customer at creation time, via `customers create --test-clock`), so attaching to an existing subscription remains a manual step in the Stripe Dashboard ("Run simulation").
+
+Get the clock id from the subscription itself: read `/v1/subscriptions/<id>` and take its `test_clock` field. If that field is null, no clock is attached yet — attach one in the Stripe Dashboard first, then advance it.
```
SaintPatrck: ❓ Does Claude have access to the Stripe Dashboard to do this?

.gitignore
```diff
@@ -32,3 +32,10 @@ pnpm-debug.log
 .claude-pr/
 review-summary.md
 review-inline-comments.md
+
+# Subagent-driven-development scratch workspace (ledgers, briefs, review packages)
+.superpowers/
```
SaintPatrck, line 37: ⚠️ If you're actively using `superpowers` visual companion (evident by ignoring this directory) please set `SUPERPOWERS_DISABLE_TELEMETRY=1` to disable their telemetry, as described [here](https://github.com/obra/superpowers#visual-companion-telemetry)

plugins/bitwarden-testing-tools/skills/using-stripe-cli/SKILL.md
```diff
@@ -0,0 +1,116 @@
...
+## Read-only debugging patterns
+
+- **Payment failures:** check the payment intent's `last_payment_error` and the latest charge's `outcome` / `failure_message` (inline it with `expand[]=latest_charge` on the payment intent).
+- **Subscription issues:** check `status`, the expanded `latest_invoice`, and recent `customer.subscription.*` events.
+- **Event tracing:** `/v1/events` filters by type, not object ID; list by type and filter client-side. Events carry the object snapshot at event time.
+- **Customer state:** expand `subscriptions` and read `--path /v1/customers/<id>/payment_methods` for the full picture.
+
+## The one permitted write: advancing an already-attached test clock
+
+You may advance a test clock that is already attached to the subscription. You may NOT attach a clock to an existing subscription: the Stripe CLI cannot do it (a test clock can only be set on a customer at creation time, via `customers create --test-clock`), so attaching to an existing subscription remains a manual step in the Stripe Dashboard ("Run simulation").
```
SaintPatrck: ⚠️ Here you say the agent "may NOT attach a clock", but the paragraph below ends with "attach one..."

plugins/bitwarden-testing-tools/skills/using-stripe-cli/SKILL.md
SaintPatrck, line 90: >— attach one in the Stripe Dashboard first, then advance it.

⚠️ This reads like an instruction to Claude, but it does not have dashboard access (that I can see). I think this should route back to the user with instructions to attach the clock then resume. Otherwise, Claude is going to fall down a rabbit hole attempting to access the dashboard itself. Something along the lines of...

> If that field is null, prompt the user with, "Attach a clock in the Stripe dashboard before continuing. Once attached, say so and I will continue."

SaintPatrck (commented):
💭 Would you be open to considering breaking the mailcatcher skill out into a separate, dedicated plugin: `bitwarden-mailcatcher-tools`?

I ask because we're currently re-evaluating the taxonomy of the marketplace and will be proposing standards and guidelines meant to improve organization to make determining where a tool should live easier for contributors, among other things. 

Making vendor integration tools their own plugin is one of those guidelines. Splitting it out now is easier than pulling it out later, but does mean the evals would need some changes.

The same guidance can technically apply to the Stripe skill, but it seems pretty tightly coupled to testing right now. I'm not sure if or how other teams would use the Stripe API.

PR #220, opened 2026-08-26
fix(bitwarden-code-review): grant git status where a local review is told to run it

withinfocus:
## 🎟️ Tracking

Layer 1 of 4 in stack #221, bottom to top: **#220**, #217, #219, #218. The first three each bump `bitwarden-code-review`, so they merge in that order.

No Jira ticket. Noticed while auditing tool grants in `bitwarden-code-review`.

## 📔 Objective

`code-review-local` tells the reviewer to inspect the working tree with `git diff` and `git status`, and its README documents the same behavior. Neither the command's `allowed-tools` nor the agent's `tools` listed `git status`, so the call prompts or fails on the local path. `performing-multi-agent-code-review` already grants it, which is why the multi-agent path works and this one does not.

Adds `Bash(git status:*)` to both.

One thing this does not fix: the agent's local mode fetches only committed work, while the command promises "uncommitted and committed changes." That is a behavior change rather than a grant fix, so it is left alone here.

Version 1.14.0 to 1.14.1.


plugins/bitwarden-code-review/agents/bitwarden-code-reviewer/AGENT.md
```diff
@@ -1,10 +1,10 @@
 ---
 name: bitwarden-code-reviewer
-version: 1.14.0
+version: 1.14.1
 description: Conducts thorough code reviews following Bitwarden standards. Finds all issues first pass, avoids false positives, respects codebase conventions. Invoke when user mentions "code review", "review code", "review", "PR", or "pull request".
 model: opus
 skills: avoiding-false-positives, classifying-review-findings, posting-bitwarden-review-comments, posting-review-summary, reviewing-dependency-changes
-tools: Read, Write, Bash(gh pr view:*), Bash(gh pr diff:*), Bash(gh pr checks:*), Bash(git show:*), Bash(gh api graphql -f query=:*), Bash(git log:*), Bash(git diff:*), Grep, Glob, Skill, mcp__github_inline_comment__create_inline_comment, mcp__github_comment__update_claude_comment
+tools: Read, Write, Bash(gh pr view:*), Bash(gh pr diff:*), Bash(gh pr checks:*), Bash(git show:*), Bash(gh api graphql -f query=:*), Bash(git log:*), Bash(git diff:*), Bash(git status:*), Grep, Glob, Skill, mcp__github_inline_comment__create_inline_comment, mcp__github_comment__update_claude_comment
```
SaintPatrck, line 7: 🎨 Can we convert this to a proper alphabetized yaml list? It makes human-parsing of this list a lot easier.

Same suggestion for the `allowed-tools` in Skills.

It's harder to overlook approved tools that way, and these are one of the more important things for humans to verify.

Not blocking, since this could be addressed at a larger scale in a ♻️ follow-up.

SaintPatrck (approved):

plugins/bitwarden-code-review/agents/bitwarden-code-reviewer/AGENT.md
withinfocus: Picked this up in #217, which was already rewriting both an agent `tools:` and a skill `allowed-tools:`. Converted to alphabetized YAML lists there: the reviewer agent, `perform-security-review`, and both `code-review` commands, with `Bash(...)` grants grouped first since those are the ones worth scanning.

I checked the form was safe before converting rather than assuming: the official `plugin-dev` frontmatter reference documents the list form for command `allowed-tools`, the shipped `humanizer` skill already uses it, and `agent-development` documents a sequence for agent `tools:`. All four files now parse as arrays under a real YAML parser.

That check also turned up something pre-existing: `argument-hint: [PR#] | [PR URL]` in both commands made the frontmatter fail to parse at all, because the unquoted `|` reads as a block-scalar indicator. It fails on `main` today. I quoted it in the same change, since a frontmatter that does not parse means these grants may never have applied.

PR #247, opened 2026-09-17
ci: verify the MCP server launch path on Node releases without Corepack

SaintPatrck:
## 🎟️ Tracking

No ticket. Follow-up to #246, which fixed the launch path; this covers it in CI. Stacked on that branch, so retarget to `main` once it merges.

## 📔 Objective

Every workflow reached pnpm through the **Corepack binary bundled with Node**, so CI depended on something recent Node releases no longer ship and never executed the launch command it was meant to cover. Nothing here would have caught the MCP server failing to connect, and nothing confirmed the repair either. The new job reads the command out of `.mcp.json` rather than restating it, runs it against a PATH holding **only node and npm**, and asserts a real MCP `initialize` response, because a process that starts without speaking MCP is exactly the failure that reached production. It is verified to exit non-zero against the previous command and zero against the current one.

The workflow's path filters now cover `.mcp.json`, which they did not, so a change to the launch command alone never triggered these tests at all.

The Corepack version is declared in the root and server manifests, beside the `packageManager` field that already appears in both, and every caller derives it from the manifest governing its own directory. That keeps the version in files **Renovate's npm manager already parses** rather than scattered across shell invocations it cannot see. The root `devEngines` declaration is dropped for the same reason the server's was, since npm refuses to run in a directory whose manifest names a different package manager with `onFail` set to error.


package.json
```diff
@@ -14,6 +14,7 @@
   },
   "homepage": "https://bitwarden.com",
   "devDependencies": {
+    "corepack": "0.36.0",
```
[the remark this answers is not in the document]
SaintPatrck: Raised with AppSec in **VULN-1072**. We're keeping the declaration here instead of on #246, because this PR is where Corepack enters the manifests and lockfiles.

.github/workflows/atlassian-mcp-server-build.yml
```diff
@@ -56,8 +56,10 @@ jobs:
 
       - name: Set up pnpm via Corepack
         run: |
-          corepack enable
-          corepack prepare --activate
+          COREPACK_VERSION=$(node -p "require('./package.json').devDependencies.corepack.trim()")
```
fedemkr, line 59: ⛏️  Would it be better to use `jq` directly here? which doesn't need node to be setup nor `trim()`. Also it'd be consistent with `PKG_VERSION` in Get package version
```
COREPACK_VERSION=$(jq -r '.devDependencies.corepack' package.json)
```

.github/workflows/atlassian-mcp-server-test.yml
```diff
@@ -50,8 +52,10 @@ jobs:
 
       - name: Set up pnpm via Corepack
         run: |
-          corepack enable
-          corepack prepare --activate
+          COREPACK_VERSION=$(node -p "require('./package.json').devDependencies.corepack.trim()")
```
fedemkr, line 55: ⛏️ Same as in `Set up pnpm via Corepack` in `*-build.yml` file.

.github/workflows/atlassian-mcp-server-test.yml
```diff
@@ -69,3 +73,76 @@ jobs:
...
+    name: Launch via .mcp.json
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        # Node 24 bundles Corepack and Node 26 does not. The launch path has to
+        # work on both without anything being installed or enabled first.
+        node-version: ["24", "26"]
```
fedemkr, line 87: ❓ Should `"22"` be taken into consideration here as well as it's one of the engines declared in the `pnpm-lock.yaml` file?

fedemkr (commented):
Looks good, just possible minor ⛏️

.github/workflows/atlassian-mcp-server-build.yml
SaintPatrck: You're right. This runs on ubuntu so we know `jq` is going to be available. Switching to `jq -er`. Including `-e` keeps the step failing fast if `corepack` is ever missing.

.github/workflows/atlassian-mcp-server-test.yml
SaintPatrck: No. `22` is `corepack`'s requirement. The MCP server declares `>=24.0.0`, so `22` is out of scope for this.

